PR #18, opened 2022-07-10
Done!

allesdmytro:
[DEMO](https://allesdmytro.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,8 +1,33 @@
...
-);
+import { Component } from 'react';
+
+type State = {
+  pressedKey: string,
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
ArtemTopolskyi, line 9: it's better to save here exactly key and render additional text directly in render method

ArtemTopolskyi (approved):

PR #29, opened 2022-07-10
Solution created

MariiaNH:
[DEMO LINK](https://mariianh.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,39 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p>Nothing was pressed yet</p>
-    <p>The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string;
+};
+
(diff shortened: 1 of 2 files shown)


ostapfriend (approved):
Good!

src/App.tsx
```diff
@@ -1,8 +1,37 @@
...
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.keyPressedEvent);
+  }
+
+  keyPressedEvent = (event: KeyboardEvent) => {
+    if (event.key) {
+      this.setState((previousState) => ({
+        ...previousState,
+        message: `The last pressed key is ${event.key}`,
+      }));
```
Olexandr6, line 25: Actually, you don't need here to use functional `setState`

Olexandr6 (approved):

src/App.tsx
Olexandr6: You don't use previous value

PR #36, opened 2022-07-11
solution

vinnibea:
[DEMO LINK](https://vinnibea.github.io/react_keyboard/)


mykola-koval (approved):
adding some value for pressing Space would be nice

PR #86, opened 2022-08-03
solution

OlgaStakhova:
[DEMO LINK](https://OlgaStakhova.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,46 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handlerKeyUp);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handlerKeyUp);
+  }
+
+  handlerKeyUp = (event: KeyboardEvent) => {
+    this.setState({
+      pressedKey: event.key,
+    });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {pressedKey
+            ? `The last pressed key is [${pressedKey}]`
+            : 'Nothing was pressed yet'}
+        </p>
+
+      </div>
+    );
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
+      </div>
+    );
+  }
+}
+
+// export const App: React.FC = () => (
+//   <div className="App">
+//     <p className="App__message">The last pressed key is [Enter]</p>
+//   </div>
+// );
```
PavloMarkov, line 46: you should delete these comments

src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
+  }
+
+  handlerKeyUp = (event: KeyboardEvent) => {
+    this.setState({
+      pressedKey: event.key,
+    });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
```
PavloMarkov, line 27: good practice! 👍

src/App.tsx
```diff
@@ -1,7 +1,46 @@
-import React from 'react';
+import { Component } from 'react';
```
PavloMarkov, line 1: 🔥 🔥 🔥 🔥 cool!

PavloMarkov (approved):
nice work!

src/App.tsx
```diff
@@ -1,7 +1,40 @@
...
+    document.addEventListener('keyup', this.handlerKeyUp);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handlerKeyUp);
+  }
+
+  handlerKeyUp = (event: KeyboardEvent) => {
+    this.setState({
+      pressedKey: event.key,
```
VETALEK, line 22: let's use `event.code` to check if the `space` is pressed and write `[space]` if so
image

VETALEK (changes requested):
GJ 👍 
commented

VETALEK (changes requested):
Need to deploy the new solution

src/App.tsx
```diff
@@ -1,7 +1,40 @@
...
+    document.addEventListener('keyup', this.handlerKeyUp);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handlerKeyUp);
+  }
+
+  handlerKeyUp = (event: KeyboardEvent) => {
+    this.setState({
+      pressedKey: event.code,
```
VETALEK, line 22: you need `event.code` to check only some special keys (like `space`), in other cases - return `event.key`
image

VETALEK (changes requested):
Almost 💪

VETALEK (approved):
👍

PR #100, opened 2022-08-04
Solution

blcts:
[DEMO LINK](https://blcts.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string
+};
+
+export class App extends React.Component< {}, State > {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.key);
+  }
+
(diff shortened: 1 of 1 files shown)


fledith (approved):
🔥

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+  componentDidMount() {
+    document.addEventListener('keyup', this.key);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.key);
+  }
+
+  key = (event: KeyboardEvent) => (
+    this.setState({ pressedKey: event.key })
```
VETALEK, line 21: let's check if the key is `space` (with `event.code`) and write `[space]` if so
image

VETALEK (changes requested):
GJ 👍 
commented

src/App.tsx
```diff
@@ -22,6 +22,10 @@ export class App extends React.Component< {}, State > {
   );
 
   render() {
+    if (this.state.pressedKey === ' ') {
```
VETALEK, line 25: better use `event.code` to check if the key is `space`
and don't forget to deploy your changes 😉 
image

VETALEK (changes requested):

Ihor-Karpyn (approved):

PR #104, opened 2022-08-04
Solution

rodiontseva:
[DEMO LINK](https://rodiontseva.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: null,
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleClick);
+  }
+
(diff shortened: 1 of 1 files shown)


roman-romanenko (approved):
Good 👍🏻 
Maybe it's no problem but you should create type for your state

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleClick);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleClick);
+  }
+
+  handleClick = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
```
VETALEK, line 21: let's write `[space]` if the `space` is pressed
image

VETALEK (changes requested):
GJ 🔥 
commented

src/App.tsx
rodiontseva: also, I have noticed that **fn** on mac os doesn't displayed

VETALEK: it's true 😢

VETALEK (approved):
🔥

PR #147, opened 2022-09-07
React keyboard implemented

andrii-dmytruk:
[DEMO LINK](https://andrii-dmytruk.github.io/react_keyboard/)


NataliyaAn (approved):
Great job!

denys-danyliuk (changes requested):
nicely done🔥

denys-danyliuk (approved):
🔥

PR #150, opened 2022-09-07
Keyboard solution

NataliyaAn:
[DEMO LINK](https://NataliyaAn.github.io/react_keyboard/)


andrii-dmytruk (approved):
Good job!

denys-danyliuk (changes requested):
good job!

denys-danyliuk (approved):
🚀

PR #152, opened 2022-09-07
Solution

VolodymyrChuyko:
[DEMO LINK](https://VolodymyrChuyko.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,38 @@
-import React from 'react';
+import { PureComponent } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string;
+};
+
+export class App extends PureComponent<{}, State> {
+  state: Readonly<State> = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener(
+      'keyup',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,38 @@
...
+  }
+
+  handleKeyup = (event: KeyboardEvent) => this.setState({ key: event.key });
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.key ? (
+            `The last pressed key is [${this.state.key}]`
```
ibondrk, line 30: You can perform destructuring of this key. It's not necessary in this case, but if you'll had more that one variable it should be good (i think you know that) )

ibondrk (approved):

src/App.tsx
```diff
@@ -1,7 +1,40 @@
...
+  state: Readonly<State> = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener(
+      'keyup',
+      this.handleKeyup,
+    );
+  }
```
denys-danyliuk, line 17:
```suggestion
  componentDidMount() {
    document.addEventListener('keyup', this.handleKeyup);
  }
```

denys-danyliuk (changes requested):
good job

denys-danyliuk (approved):
🚀

PR #154, opened 2022-09-08
Solution

adrqq:
[DEMO LINK](https://adrqq.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,36 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount(): void {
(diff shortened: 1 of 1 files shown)


OleksiiPustovit (changes requested):
The code looks very good, but the site on the demo link does not open, try to repeat the deploy.

adrqq:
It should work now

OleksiiPustovit (approved):
Good job 🔥

src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {`${pressedKey.length !== 0
```
denys-danyliuk, line 29:
```suggestion
          {pressedKey.length !== 0
```

denys-danyliuk (changes requested):
GJ

denys-danyliuk (changes requested):
you forgot to deploy🥲

adrqq:
Now should be right

denys-danyliuk (approved):
🚀

PR #158, opened 2022-09-08
react keyboard solution

artem-kodesnikov:
[DEMO LINK](https://artem-kodesnikov.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,39 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    isPressed: true,
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    this.state.isPressed = false;
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
+  componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    isPressed: true,
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    this.state.isPressed = false;
```
dellachaise, line 14: This code is redundand.

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
```
dellachaise, line 16: You duplicate the callback function. Write it separately and use in 2 methods.

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+    });
+  }
+
+  render() {
+    const { pressedKey, isPressed } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {isPressed
```
dellachaise, line 32: You could simply use here `pressedKey`.
```suggestion
          {pressedKey
```

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+            ? 'Nothing was pressed yet'
+            : `The last pressed key is [${pressedKey}]`}
+        </p>
+      </div>
+    );
+  }
+}
```
dellachaise, line 39: Optional: would be better use event.code to check if the pressed key is space and write [space] if so

dellachaise (changes requested):
Great start 👍 

dellachaise (approved):
Great job 👍

PR #167, opened 2022-09-09
solution 1

kostya-kosyuk:
[DEMO LINK](https://kostya-kosyuk.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.eventPressKey);
+  }
+
+  componentWillUnmount(): void {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.pressedKey !== ''
+            ? `The last pressed key is [${this.state.pressedKey}]`
+            : 'Nothing was pressed yet'}
+        </p>
+      </div>
```
SvyatoslavYatsenko, line 32:
```suggestion
  render() {
  const { pressedKey } = this.state;
    return (
      <div className="App">
        <p className="App__message">
          {this.state.pressedKey !== ''
            ? `The last pressed key is [${pressedKey}]`
            : 'Nothing was pressed yet'}
        </p>
      </div>
```

SvyatoslavYatsenko (approved):
Well done!

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  handlePressedKey = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.pressedKey !== ''
```
dellachaise, line 28: `this.state.pressedKey !== '' ` converts to true and if we have  pressedKey we also always get true. This condition would be enough 👇 
```suggestion
          {this.state.pressedKey
```

dellachaise (approved):
Well done 👍 

PR #170, opened 2022-09-09
Solution

DimaKobzar7:
[DEMO LINK](https://DimaKobzar7.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -17,17 +17,11 @@ export class App extends Component<{}, State> {
     document.removeEventListener('keyup', this.handleKeyup);
   }
 
-  handleKeyup = (event: KeyboardEvent) => {
-    if (event.code === 'Space') {
-      this.setState({
-        pressedKey: event.code,
-      });
-    } else {
-      this.setState({
-        pressedKey: event.key,
-      });
-    }
-  };
+  handleKeyup = (event: KeyboardEvent) => (
+    event.code === 'Space'
+      ? this.setState({ pressedKey: event.code })
+      : this.setState({ pressedKey: event.key })
+  );
 
   render() {
     const { pressedKey } = this.state;


adrqq (approved):
Good job, but to s1mplify your code remove pressedKey and just check if the length of 
keyBoardBtn > 0

DimaKobzar7:
pressedKey  is the task condition.
Make the App a class component with pressedKey in the state.

adrqq:
Didn't see it, you're right.

src/App.tsx
```diff
@@ -1,7 +1,39 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: boolean,
```
dellachaise, line 4: Why do you need this variable? Try to rewrite without it.

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+export class App extends Component<{}, State> {
+  state: State = {
+    pressedKey: false,
+    keyBoardBtn: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: true, keyBoardBtn: event.key });
+    });
```
dellachaise, line 17: You write here and inlines 21-22 almost the same code. When you would delete `pressesKey` variable. You could create a separate function for this purpose and use it in 2 places.

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+  }
+
+  render() {
+    const { pressedKey, keyBoardBtn } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {!pressedKey
```
dellachaise, line 32: You could just use here `keyBoardBtn`.

dellachaise (changes requested):
Great start 👍 

src/App.tsx
DimaKobzar7: pressedKey is the task condition.
Make the App a class component with pressedKey in the state.

dellachaise: Yes, but you could hold there pressedKey, not boolean.

src/App.tsx
```diff
@@ -1,7 +1,45 @@
...
+  handleKeyup = (event: KeyboardEvent) => {
+    if (event.code === 'Space') {
+      this.setState({
+        pressedKey: event.code,
+      });
+    } else {
+      this.setState({
+        pressedKey: event.key,
+      });
+    }
```
dellachaise, line 29: Better use ternary operator here.

dellachaise (approved):
Well done 👍

PR #171, opened 2022-09-09
Solution

KorniychukLesya:
[DEMO LINK](https://korniychuklesya.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  curentKey: string,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    curentKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ curentKey: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ curentKey: event.key });
+    });
+  }
+
+  render() {
+    const { curentKey } = this.state;
+
+    return (
+      <div className="App">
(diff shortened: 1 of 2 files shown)


anna-kolmakova (approved):
Good job!)

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  curentKey: string,
+};
+
+export class App extends React.Component<{}, State> {
```

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  state = {
+    curentKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ curentKey: event.key });
```
dellachaise, line 14: Do not duplicate code(here and lines 19-20). Create a separate function and use it.

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    });
+  }
+
+  render() {
+    const { curentKey } = this.state;
+
+    return (
+      <div className="App">
+        {(curentKey === ''
```
dellachaise, line 29: This condition would be enough,  use  `!` here.
```suggestion
        {(!curentKey
```

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+          ? <p className="App__message">Nothing was pressed yet</p>
+          : <p className="App__message">{`The last pressed key is [${curentKey}]`}</p>
+        )}
+
+      </div>
+    );
+  }
+}
```
dellachaise, line 37: Optional: would be better use event.code to check if the pressed key is space and write [space] if so

dellachaise (changes requested):
Great start 👍 

KorniychukLesya:
Thanks for checking! 
I fixed everything)

dellachaise (approved):
Well done 👍

PR #177, opened 2022-09-10
keyboard logger solution

ibondrk:
[DEMO LINK](https://ibondrk.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,41 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  lastPressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    lastPressedKey: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', (this.pressedKey));
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (this.pressedKey));
+  }
+
+  pressedKey = (event: KeyboardEvent): void => {
+    this.setState({
+      lastPressedKey: event.key,
+    });
+  };
+
+  render() {
+    const { lastPressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+  lastPressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    lastPressedKey: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', (this.getPressedKey));
```
VolodymyrChuyko, line 13:
```suggestion
    document.addEventListener('keyup', this.getPressedKey);
```

src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (this.getPressedKey));
```
VolodymyrChuyko, line 17:
```suggestion
    document.removeEventListener('keyup', this.getPressedKey);
```

VolodymyrChuyko (approved):
Good job!

src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+type State = {
+  lastPressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    lastPressedKey: '',
+  };
+
+  componentDidMount(): void {
```
denys-danyliuk, line 12:
```suggestion
  componentDidMount() {
```

denys-danyliuk (changes requested):
good job

src/App.tsx
ibondrk: I understand what you're talking about, but if I change `event.key` to `event.code`, the rest of the buttons will be displayed otherwise, so I added a condition to the code so that space is displayed

denys-danyliuk: Yes, this too will work, you even could do this check inside getPressedKey.
It does actually, probably should set is as optional, but you've done a great job. You just forgot to deploy your changes🥲

denys-danyliuk (approved):
🚀

src/App.tsx
ibondrk: You're right, I forgot😅. Already did it

PR #182, opened 2022-09-11
react keyboard

anyakrainiuk:
[DEMO LINK](https://anyakrainiuk.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    message: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.pressedKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.pressedKey);
+  }
+
+  pressedKey = (event: KeyboardEvent) => {
+    this.setState({
+      message: `The last pressed key is [${event.key}]`,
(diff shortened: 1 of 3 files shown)


YuriiPronin (approved):
Nice work!

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string;
+};
+
+export class App extends Component<{}, State> {
```

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+  }
+
+  keyPressed = (event: KeyboardEvent) => {
+    this.setState({
+      message: `The last pressed key is [${event.key}]`,
+    });
+  };
+
+  render() {
+    // const { message } = this.state;
```
dellachaise, line 27: Better use destructuring.

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+    return (
+      <div className="App">
+        <p className="App__message">{this.state.message}</p>
+      </div>
+    );
+  }
+}
```
dellachaise, line 35: Optional: would be better use event.code to check if the pressed key is space and write [space] if so

dellachaise (changes requested):
Great job 👍 

src/App.tsx
```diff
@@ -1,7 +1,43 @@
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.keyPressed);
+  }
+
+  keyPressed = (event: KeyboardEvent) => {
+    if (event.key === 'space') {
+      this.setState({
+        message: 'The last pressed key is [ ]',
```
denys-danyliuk, line 25:
```suggestion
        message: 'The last pressed key is [space]',
```

denys-danyliuk (approved):

PR #185, opened 2022-09-12
React Keyboard was implemented

Anton-Liada:
[DEMO LINK](https://Anton-Liada.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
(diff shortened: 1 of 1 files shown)


Sasha-Gordijchuk (approved):
🔥 🔥 🔥

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
```

dellachaise (approved):
Well done 👍 

Optional: would be better use `event.code` to check if the pressed key is space and write [space] if so

PR #186, opened 2022-09-12
Solution

vladgusakovskiy:
[DEMO LINK](https://vladgusakovskiy.github.io/react_keyboard/)


roma-shafihullin (approved):
good solution 👍

denys-danyliuk (changes requested):
GJ
let's use event.code to check if the pressed key is space and write [spac
e] if so

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleDocumentEvent);
+  }
+
+  handleDocumentEvent = (event: KeyboardEvent) => {
+    if (event.key === 'space') {
+      this.setState({ pressedKey: 'The last pressed key is [space]' });
+    } else {
+      this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
```
denys-danyliuk, line 24: don't save a whole string to pressedKey, save only a name of the key

denys-danyliuk (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,43 @@
...
+
+  handleDocumentEvent = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+
+    if (this.state.pressedKey === ' ') {
+      this.state.pressedKey = 'space';
+    }
+  };
+
+  render(): React.ReactNode {
```
denys-danyliuk, line 28:
```suggestion
  render() {
```

denys-danyliuk (approved):
Good job, you probably forgot to do deploy, because when I press space I see empty space

PR #189, opened 2022-09-12
Develop

OlgaOchichenko:
[DEMO LINK](https://OlgaOchichenko.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string,
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    message: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.pressedKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.pressedKey);
+  }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+  state: Readonly<State> = {
+    message: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.pressedKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.pressedKey);
```
Jamal-Amoian, line 17: if you added "keyup", why you removed "keydown?"

Jamal-Amoian (approved):
Well done!

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string,
+};
+
+export class App extends Component<{}, State> {
```

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.pressedKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.pressedKey);
+  }
+
+  pressedKey = (event: KeyboardEvent) => {
```
dellachaise, line 20: The name of functions should start with **verb.**
```suggestion
  pressKey = (event: KeyboardEvent) => {
```

dellachaise (approved):
Well done 👍 

PR #205, opened 2022-09-15
react keyboard solution

SergeyPistelev:
[DEMO LINK](https://sergeypistelev.github.io/react_keyboard/)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
```
yevhenii-podplietnii, line 22: I think you should create separate method instead of creating inline event handlers

yevhenii-podplietnii (approved):
Good job

AlinaMatuschak (changes requested):
👍
just fix @yevhenii-podplietnii comment

src/App.tsx
SergeyPistelev: thanks for the comment, method has been added

yevhenii-podplietnii (approved):

AlinaMatuschak (approved):

PR #223, opened 2022-10-03
solution

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string | null,
+  callback: Function,
+}
+
+export class App extends React.Component<{}, State>{
+  state = {
+    key: null,
+    callback: (event: KeyboardEvent) => {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-);
+type State = {
+  key: string | null,
+  callback: Function,
+}
+
+export class App extends React.Component<{}, State>{
+  state = {
+    key: null,
+    callback: (event: KeyboardEvent) => {
```
ihorandrianov, line 11: we pass to state data that can be changed. Callback function never changes so you don't need it in state

ihorandrianov (commented):
Nice work!

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string | null,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    key: null,
```
OleksiiBlezniuk, line 9: usually we set value of same type with falsy value here

for example empty string, no need to set null for every type

OleksiiBlezniuk (approved):

PR #224, opened 2022-10-04
solution

ihorandrianov:
[DEMO LINK](https://ihorandrianov.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,47 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string | null
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: null,
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (e) => {
+      this.setState({
+        key: e.key,
+      });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (e) => {
+      this.setState({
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,47 @@
...
+      this.setState({
+        key: e.key,
+      });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (e) => {
+      this.setState({
+        key: e.key,
```
MaxManis, line 23: this way of removing of event listener wont work, because you passed a completely new function)

src/App.tsx
```diff
@@ -1,7 +1,47 @@
...
+            ? (
+              <div>
+                {`The last pressed key is ${this.state.key}`}
+              </div>
+            )
+            : (
+              <div>
+                Nothing was pressed yet
+              </div>
+            )
```
MaxManis, line 42: you also could use this ternaty statement inside of your div since its doesnt change)

src/App.tsx
```diff
@@ -1,7 +1,45 @@
...
+        }
```
OleksiiBlezniuk, line 41: Div's are same, just choose needed string by ternary operator

src/App.tsx
```diff
@@ -1,7 +1,45 @@
...
+    this.setState({
+      key: e.key,
+    });
+  };
+
+  render() {
+    return (
+      <>
+        {
+          this.state.key
```
OleksiiBlezniuk, line 30: Would be nice to use destructuring for state

OleksiiBlezniuk (changes requested):

silchyn (approved):
Cool!

PR #237, opened 2022-10-13
Solution

Apollonia5:
[DEMO LINK](https://Apollonia5.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,42 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: boolean;
+  keyName: string;
+};
+
+export class App extends Component<{}, State> {
+  state: State = {
+    pressedKey: false,
+    keyName: '',
+  };
+
+  componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,42 @@
...
-  </div>
-);
+type State = {
+  pressedKey: boolean;
+  keyName: string;
+};
+
+export class App extends Component<{}, State> {
+  state: State = {
+    pressedKey: false,
```
OleksiiBlezniuk, line 10: why do you need this state?
just save keyName and render it if exists

src/App.tsx
```diff
@@ -1,7 +1,42 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keydown', this.pressKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.pressKey);
+  }
+
+  pressKey = (event: KeyboardEvent) => {
```
OleksiiBlezniuk, line 22:
```suggestion
  handleKeyPress = (event: KeyboardEvent) => {
```

OleksiiBlezniuk: Ususally we use `handle` or `handler` word in event handler functions

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

PR #239, opened 2022-10-13
Develop

vladimir-dublya:
https://vladimir-dublya.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string | null
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: null,
+  };
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    });
+  };
+
+  render() {
+    const { key } = this.state;
+
+    return (
+      <div>
+        {key
+          ? `The last pressed key is ${key}`
```
Andrew-Skoryk, line 32: `The last pressed key is [${key}]`

Andrew-Skoryk (approved):
GJ! I Would recommend changing the demo link and adding '[]' to fix tests results

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string | null
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: null,
```
Anna-Tkachenko, line 9: will be better to use empty string here instead of null

Anna-Tkachenko (approved):

PR #242, opened 2022-10-13
Component keyboard has been created

iam-datsenko:
[DEMO LINK](https://iam-datsenko.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,14 +10,14 @@ export class App extends Component<{}, State> {
   };
 
   componentDidMount() {
-    document.addEventListener('keyup', this.hendleKeyboard);
+    document.addEventListener('keyup', this.handleKeyboard);
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


OleksandrGlulhovskiy (commented):
GJ!

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ key: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ key: event.key });
+    });
```
Anna-Tkachenko, line 21: it won't work
you should create separate function as listener, and add and delete the same function in both of methods

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #249, opened 2022-10-14
created App component with keyboard click handler

ArtemHulkoDev:
[DEMO LINK](https://ArtemHulkoDev.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,29 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.pressedKeyHandler);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.pressedKeyHandler);
+  }
+
+  pressedKeyHandler = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">{pressedKey}</p>
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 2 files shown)


xiimiik (approved):
Greate!

src/App.tsx
```diff
@@ -1,7 +1,29 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
Anna-Tkachenko, line 5: you should store only key value in state, not full text message

Anna-Tkachenko (changes requested):

src/App.tsx
ArtemHulkoDev: fixed)

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: null,
+  };
```
Anna-Tkachenko, line 6: create type for the state

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {(pressedKey && `The last pressed key is [${pressedKey}]`) || 'Nothing was pressed yet'}
```
Anna-Tkachenko, line 26: use ternary operator here

Anna-Tkachenko (changes requested):

src/App.tsx
ArtemHulkoDev: done

src/App.tsx
ArtemHulkoDev: done

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | null;
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: null,
```
Anna-Tkachenko, line 9: will be better to use empty string here

Anna-Tkachenko (approved):
⭐️

PR #250, opened 2022-10-14
keyboard was implemented

dsydorenko534:
[DEMO LINK](https://dsydorenko534.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,39 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string,
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
+  };
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.Pressedkey);
+  }
+
+  Pressedkey = (event: KeyboardEvent) => {
+    if (event.key) {
+      this.setState({ key: event.key });
+    }
+  };
```
iPanda177, line 24: you don't need `if` here

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
+            this.state.key
+              ? `The last pressed key is [${this.state.key}]`
```
iPanda177, line 32: use destructuring and put value into a variable

iPanda177 (changes requested):
Cool that you added some styles!)
Left you some small comments

PR #274, opened 2022-10-17
added RCC with key listener

igor-1987:
[DEMO LINK](https://igor-1987.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string
+};
+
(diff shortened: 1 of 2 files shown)


xiimiik (approved):
Greate!

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
Anna-Tkachenko, line 9: you should store only key value in state, not full message

Anna-Tkachenko (changes requested):

src/App.tsx
igor-1987: Ok, done.

Anna-Tkachenko (approved):
⭐️

PR #281, opened 2022-10-17
react keyboard

Arsen-Shpak:
[DEMO LINK](https://Arsen-Shpak.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  render() {
+    const { key } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {key === ''
+            ? 'Nothing was pressed yet'
+            : `The last pressed key is [${key}]`}
```
igor-1987, line 32: I guess the better way is to manage this text depending directly from on the state value, not to make additional statement.

igor-1987 (approved):

src/App.tsx
OleksiiBlezniuk, line 32:
```suggestion
          {!key
            ? 'Nothing was pressed yet'
            : `The last pressed key is [${key}]`}
```

OleksiiBlezniuk (approved):

PR #287, opened 2022-10-18
solution

Yaroslav-Radchuk:
[DEMO LINK](https://Yaroslav-Radchuk.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,39 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string,
```
tetiana-ivanchenko, line 4: if here you don't save key, but some message, it would be better raname this property more explicit

tetiana-ivanchenko (approved):

PR #294, opened 2022-10-21
react_keyboard

Blenemy:
[DEMO LINK](https://Blenemy.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,36 @@
+// import { prependOnceListener } from 'process';
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string,
+};
+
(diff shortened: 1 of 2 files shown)


loralevitska (changes requested):
good job! almost done. But your solution have some difference between example working version. For example try to use `Tab` on your app and [working version](https://mate-academy.github.io/react_keyboard/)

fix your handler
image

src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleEvent);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleEvent);
+  }
+
+  handleEvent = (event: KeyboardEvent) => {
```
polosanya, line 21: Better to clarify which event

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment

PR #295, opened 2022-10-21
react_keyboard

zhenija:
[DEMO LINK](https://zhenija.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.pressKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.pressKey);
+  }
+
+  pressKey = ((event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  });
+
+  render() {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    document.addEventListener('keyup', this.pressKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.pressKey);
+  }
+
+  pressKey = ((event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  });
```
polosanya, line 22: Don't use a function before declaration. Move it above life cycle methods.
Also please change the name, because for now `pressKey` and `pressedKey` look almost the same and can be easily confused.
```suggestion
  handlePressKey = ((event: KeyboardEvent) => {
    this.setState({ pressedKey: event.key });
  });
```

polosanya (changes requested):
Good job! Please check my comments that may help you to improve your code)

src/App.tsx
zhenija: I fixed it, thanks! But with function i had to add an 'eslint-disable-next-line'.
image

src/App.tsx
```diff
@@ -1,7 +1,38 @@
...
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  // eslint-disable-next-line react/sort-comp
+  handlePressKey = ((event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  });
```
polosanya, line 15:
```suggestion
  handlePressKey = (event: KeyboardEvent) => {
    this.setState({ pressedKey: event.key });
  };
```

polosanya (changes requested):
Eslint tells you to move your funciton below componentDidUnmount.
Don't ignore it's advices)
image

PR #302, opened 2022-10-30
Solution

Viktorianeimesh:
[DEMO LINK](https://Viktorianeimesh.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends Component {
+  state = {
+    pressedKey: false,
+    keySymbol: '',
+  };
+
+  componentDidMount() {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    }
+  };
+
+  render() {
+    const { pressedKey, keySymbol } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {pressedKey
```
polosanya, line 30: You can use your keySymbol instead.
If any key was pressed, your keySymbol will have some value, otherwise it will be empty string

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends Component {
+  state = {
+    pressedKey: false,
```
polosanya, line 5: I think this variable is redundant

polosanya (changes requested):
Almost perfect! Please check my comments

polosanya (approved):
Well done!

PR #308, opened 2022-11-09
Solution

Tanya-Grebenkina:
[DEMO LINK](https://Tanya-Grebenkina.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,14 +10,14 @@ export class App extends Component<{}, State> {
   };
 
   componentDidMount() {
-    document.addEventListener('keyup', this.handlerKeyEvent);
+    document.addEventListener('keyup', this.keyEventHandler);
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handlerKeyEvent);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handlerKeyEvent);
+  }
+
+  handlerKeyEvent = (event: KeyboardEvent) => {
```
polosanya, line 20:
```suggestion
  keyEventHandler = (event: KeyboardEvent) => {
```
It is a good tone not to use function before declaration. Please move it above
Also, I think keyEventHandler would sound better

polosanya (changes requested):
Almost perfect! Please pay attention to my comments

liliia-livandovskaya (approved):

PR #309, opened 2022-11-09
React keyboard task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedButton: string;
+};
+
+export class App extends React.Component {
+  state: Readonly<State> = {
+    pressedButton: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keydown', (e) => {
+      const pressedButton = e.key;
+
+      this.setState({ pressedButton });
+    });
+  }
+
+  render() {
+    const { pressedButton } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          { pressedButton ? `The last pressed key is [${pressedButton}]` : 'Nothing was pressed yet'}
+        </p>
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+};
+
+export class App extends React.Component {
+  state: Readonly<State> = {
+    pressedButton: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keydown', (e) => {
+      const pressedButton = e.key;
```
polosanya, line 14:
```suggestion
      const newPressedButton = e.key;
```
For now your name is a bit confusing with state variable

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+    });
+  }
+
+  render() {
+    const { pressedButton } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          { pressedButton ? `The last pressed key is [${pressedButton}]` : 'Nothing was pressed yet'}
```
polosanya, line 26:
```suggestion
          { pressedButton 
            ? `The last pressed key is [${pressedButton}]` 
            : 'Nothing was pressed yet'
          }
```

src/App.tsx
```diff
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedButton: string;
```
polosanya, line 4: Btw, it should be called pressedKey
image
```suggestion
  pressedKey: string;
```

polosanya (changes requested):
Good job! Please check my comments

src/App.tsx
juliiaap: renamed

src/App.tsx
juliiaap: fixed

src/App.tsx
juliiaap: fixed

liliia-livandovskaya: this one is not fixed

liliia-livandovskaya (approved):

src/App.tsx
juliiaap: state variable was renamed as `pressedKey`, thus state variable and const in componentDidMount() have a different name

PR #319, opened 2022-11-17
Solution

viktor-kotik:
[DEMO LINK](https://<your_account>.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,36 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


qoa11a (changes requested):
Demo link doesn't work. Let's fix it.

viktor-kotik:
Ok, sorry

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', () => {
+      this.setState({ pressedKey: '' });
+    });
```
qoa11a, line 17: It won't remove event listener. You need to store function and pass it in addEventListener and in removeEventListener. That's because functions are objects and are stored by reference rather than by value.

qoa11a (changes requested):

src/App.tsx
viktor-kotik: Ok, thanks

loralevitska (approved):
good job!

PR #321, opened 2022-11-17
keboard

sh1gatsu:
DEMO LINK: https://sh1gatsu.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,43 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (this.getEventOfGlobalHandler));
+  }
+
+  componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


MaxFedak (approved):
good job

src/App.tsx
```diff
@@ -1,7 +1,43 @@
...
+          </p>
+        </div>
+      )
+      : (
+        <div className="App">
+          <p className="App__message">
+            Nothing was pressed yet
+          </p>
+        </div>
+      );
```
denys-danyliuk, line 41:
```suggestion
    return
        <div className="App">
          <p className="App__message">
            {pressedKey 
              ? `The last pressed key is [${pressedKey}]`
              :   Nothing was pressed yet
          </p>
        </div>
      )
```

denys-danyliuk: it's the same block and you just need to render different text so there is no need in re-rendering the whole block

src/App.tsx
```diff
@@ -1,7 +1,43 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', (this.getEventOfGlobalHandler));
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (this.getEventOfGlobalHandler));
+  }
+
+  getEventOfGlobalHandler = (event: KeyboardEvent) => {
```
denys-danyliuk, line 20: try to find more representative name, now it's unclear what does it do

denys-danyliuk (changes requested):

denys-danyliuk (approved):

PR #326, opened 2022-11-18
solution

mitattt:
[DEMO LINK](https://mitattt.github.io/react_keyboard/)

@Olexandr6 cant implement your advice, tests are failing

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,16 +10,16 @@ export class App extends Component<{}, State> {
   };
 
   componentDidMount() {
-    document.addEventListener('keyup', (event: KeyboardEvent) => {
-      this.setState({ key: event.key });
-    });
+    document.addEventListener('keyup', this.getKey);
   }
 
   componentWillUnmount() {
-    document.removeEventListener('keyup', (event: KeyboardEvent) => {
-      this.setState({ key: event.key });
-    });
+    document.removeEventListener('keyup', this.getKey);
   }
+
+  getKey = (event: KeyboardEvent) => {
+    this.setState({ key: event.key });
+  };
 
   render() {
     const { key } = this.state;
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ key: event.key });
+    });
+  }
```
Svitlana-Yudina, line 16: Maybe better save this function to a variable) You repeat the same function twice)

Svitlana-Yudina (approved):
Great Job as for me)) 🔥

src/App.tsx
mitattt: ok

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.getKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.getKey);
+  }
+
+  getKey = (event: KeyboardEvent) => {
```
Olexandr6, line 20: `getKey` - means the functions must return key, but it returns nothing)
Better to name - handleKeyPress

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
```
Olexandr6, line 9: You can put here: `Nothing was pressed yet`, in this case, you won't need to conditional rendering in your render method

Olexandr6 (approved):

PR #338, opened 2022-11-20
Develop

taraselnik:
[DEMO LINK](https://taraselnik.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,49 @@
 import React from 'react';
+// import classNames from 'classnames';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type Props = {
+};
+
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<Props, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keydown', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
+  render() {
+    const { pressedKey } = this.state;
+
+    // pressedKey = 'AAA';
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {pressedKey
+            ? `The last pressed key is [${pressedKey}]`
+            : 'Nothing was pressed yet'}
+        </p>
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,49 @@
...
-);
+type Props = {
+};
+
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<Props, State> {
+  state = {
```
Svitlana-Yudina, line 12: I think, it's not necessary to create empty Props. You can use just {} in generic))
```suggestion
type State = {
  pressedKey: string;
};

export class App extends React.Component<{}, State> {
  state = {
```

src/App.tsx
```diff
@@ -1,7 +1,49 @@
...
+    pressedKey: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keydown', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
```
Svitlana-Yudina, line 20: Better move this function to the separate variable, because you repeat it two times))

src/App.tsx
```diff
@@ -1,7 +1,49 @@
...
+      </div>
+    );
+  }
+}
+
+// export const App2: React.FC = () => (
+//   <div className="App">
+//     <p className="App__message">The last pressed key is [Enter]</p>
+//   </div>
+// );
```
Svitlana-Yudina, line 49: and remove comments 😃

Svitlana-Yudina (approved):
Great Job)) Just little remarks)

src/App.tsx
denys-danyliuk: and because you don't actually delete it in componentWillUnmount
https://mate.academy/learn/react/class-components#/video/3221
watch from 5:05

denys-danyliuk (changes requested):

src/App.tsx
taraselnik: ok

src/App.tsx
taraselnik: ok

src/App.tsx
taraselnik: ook

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  componentDidMount(): void {
+    document.addEventListener('keydown', this.handleDocumentKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.handleDocumentKey);
+  }
+
+  handleDocumentKey = (event: KeyboardEvent) => {
```
denys-danyliuk, line 20:
```suggestion
  handleKeyPress = (event: KeyboardEvent) => {
```

denys-danyliuk (approved):

PR #349, opened 2022-11-21
Keyboard is ready

shtht24:
[DEMO LINK](https://shtht24.github.io/react_keyboard/)


MikhailKucherina (changes requested):
it work's good, but maybe it will be better to print 'Space', when it pressed, insted of empty string :) try to add it

shtht24:
> it work's good, but maybe it will be better to print 'Space', when it pressed, insted of empty string :) try to add it

MikhailKucherina (approved):

denys-danyliuk (approved):

denys-danyliuk (approved):

PR #367, opened 2022-12-02
Added the task solution

oleksusov:
[DEMO LINK](https://oleksusov.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,48 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  hasPressedKey: boolean,
+  pressedKey: string,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    hasPressedKey: false,
+    pressedKey: '',
+  };
+
+  componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,48 @@
...
+
+        {hasPressedKey ? (
+          <p className="App__message">
+            {`The last pressed key is [${pressedKey}]`}
+          </p>
+        ) : (
+          <p className="App__message">
+            Nothing was pressed yet
+          </p>
+        )}
```
loralevitska, line 43: You have a duplicate wrapper to the text, transfer the condition to the text only

loralevitska (changes requested):
almost perfect)

src/App.tsx
```diff
@@ -1,7 +1,42 @@
...
-);
+type State = {
+  hasPressedKey: boolean,
+  pressedKey: string,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    hasPressedKey: false,
+    pressedKey: '',
```
polosanya, line 11: Do you really need both of them? 
If `pressedKey` is not an empty string, then apparently some key was pressed 🤔

polosanya (changes requested):
Let's polish it up)

liliia-livandovskaya (approved):

PR #372, opened 2022-12-14
Keyboard

s3rhiienko:
[DEMO LINK](https://s3rhiienko.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,7 +6,7 @@ type State = {
 
 export class App extends Component<{}, State> {
   state = {
-    pressedKey: 'Nothing was pressed yet',
+    pressedKey: '',
   };
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string,
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
Bohdan-Maksymiuk, line 9:
```suggestion
    pressedKey: 'Nothing was pressed yet',
```
'Nothing was pressed yet' - is not a key
better to store only key in state and then compute string for user in `render()`

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #373, opened 2022-12-14
Solution

shyptia:
[DEMO LINK](https://shyptia.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handler);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handler);
+  }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handler);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handler);
+  }
+
+  handler = (event: KeyboardEvent) => {
```
solaryasha, line 20: Method's name should start with a verb (so you could easily tell what your method actually do)
```suggestion
  handleKeyPress = (event: KeyboardEvent) => {
```

solaryasha: Let's render `space` if I clicked the Space button
image

solaryasha (changes requested):

src/App.tsx
```diff
@@ -10,15 +10,19 @@ export class App extends React.Component<{}, State> {
...
 
   componentWillUnmount() {
-    document.removeEventListener('keyup', this.handler);
+    document.removeEventListener('keyup', this.handleKeyPress);
   }
 
-  handler = (event: KeyboardEvent) => {
-    this.setState({ pressedKey: event.key });
+  handleKeyPress = (event: KeyboardEvent) => {
+    if (event.code === 'Space') {
```
solaryasha, line 21: It's better to compare with some enum values.
```
enum KeybordKey {
  Space = 'Space'
}

if (event.code === KeybordKey.Space) {
  // some logic
}

solaryasha (approved):

PR #403, opened 2022-12-28
[DEMO LINK]

batstolya:
[DEMO LINK](https://batstolya.github.io/react_keyboard/)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({
+        pressedKey: `The last pressed key is [${event.key}]`,
+      });
+    });
+  }
```
romanblyashuk94, line 26: it doesnt work :)

```suggestion
handlePressedKey = (event) => ...

  componentDidMount() {
    document.addEventListener('keyup', handlePressedKey);
  }

  componentWillUnmount() {
    document.removeEventListener('keyup', handlePressedKey);
  }
```

romanblyashuk94 (changes requested):
Nice styles 👍  :)

src/App.tsx
batstolya: @romanblyashuk94  Thanks for the explanation ! I fixed it

romanblyashuk94 (approved):
great 👍

valerii-derkach (approved):
🔥

PR #411, opened 2022-12-29
solution

APetrinko:
[DEMO LINK](https://APetrinko.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,36 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string;
+};
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+        {
+          this.state.key
+          // eslint-disable-next-line
+            ? <p className="App__message">The last pressed key is{`[${this.state.key}]`}</p>
+            : <p>Nothing was pressed yet</p>
+        }
+      </div>
+    );
+  }
+}
```
romanblyashuk94, line 36: the tests expect that element p is always present
thats why it fails
```suggestion
		const { key } = this.state;
		
    return (
      <div className="App">
        <p className="App__message">
          {key
            ? `The last pressed key is [${key}]`
            : 'Nothing was pressed yet'}
        </p>
      </div>
    );
  }
}
```

romanblyashuk94 (changes requested):
And add some styles

PR #416, opened 2022-12-29
in progress

OlggaJul:


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+      this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
+    });
+  }
+
```
RedkoSerhii, line 23: you have the same body in Did&Will Unmount, will be better use function

RedkoSerhii (changes requested):

src/App.tsx
OlggaJul: 👌🏻

RedkoSerhii (approved):
Great job!

src/App.tsx
```diff
@@ -1,7 +1,33 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
```
Ihor-Karpyn, line 4: a variable called 'pressedKey' but u save here message

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):

PR #423, opened 2022-12-29
solution

korchovskyi:
- [DEMO LINK](https://korchovskyi.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string | null;
+}
+
+export class App extends React.Component<{}, State> {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ key: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ key: event.key });
+    });
+  }
```
romanblyashuk94, line 22: it doesnt work :)

romanblyashuk94 (changes requested):

src/App.tsx
korchovskyi: fixed, thank you!

romanblyashuk94 (approved):
Good job 👍

solaryasha (approved):
Nice!

PR #427, opened 2022-12-29
keybord

victoria-kovalenko:
[DEMO LINK](https://victoria-kovalenko.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string
+};
+
+export class App extends Component<{}, State> {
+  state:Readonly<State> = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      if (event.key) {
+        this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
+      }
+    });
+  }
```
romanblyashuk94, line 26: it doesnt work :)

romanblyashuk94 (changes requested):

romanblyashuk94 (approved):
exelent :)

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | null,
+};
+
+export class App extends Component<{}, State> {
+  state:Readonly<State> = {
```
solaryasha, line 8:
```suggestion
  state: Readonly<State> = {
```

PR #436, opened 2022-12-30
Solution

sercansimsek:


src/App.tsx
```diff
@@ -1,7 +1,40 @@
...
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type Props = {};
+type State = {
+  pressedKey: string
+};
+
+export class App extends React.Component<Props, State> {
```
nurdanokcu, line 8:
```suggestion
export class App extends React.Component<{}, State> {
```
You can just put empty object here and remove the `type Props = {};` line

src/App.tsx
```diff
@@ -1,7 +1,40 @@
...
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
```
nurdanokcu, line 17:
```suggestion
  componentDidMount() {
    document.addEventListener('keyup', this.handleKey);
  }
  
  handleKey = (event: KeyboardEvent) => {
      this.setState({ pressedKey: event.key })
   }
```
This way , the code block wouldn't be repeated twice.

nurdanokcu (approved):
Amazing work.  
I added my suggestions but they are minor things.

meforoma (changes requested):
keypress handler is totally to be declared separately for DRY sake

meforoma (approved):
🎉

PR #439, opened 2022-12-31
done

IoghiStudio:
[DEMO LINK](https://IoghiStudio.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,47 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type Props = {};
+
+type State = {
+  key: string;
+};
+
+export class App extends React.Component<Props, State> {
+  state = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKeyEvent);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyEvent);
+  }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,47 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type Props = {};
+
+type State = {
+  key: string;
+};
+
+export class App extends React.Component<Props, State> {
```
dolnaro, line 9: you don't have to put `type Props` in this case because there is no type definition (If I am not mistaken), you just can put `<{}, State>` on line 9.

src/App.tsx
```diff
@@ -1,7 +1,47 @@
...
+    this.setState({
+      key: event.key,
+    });
+
+    if (event.key === ' ') {
+      this.setState({
+        key: 'Space',
+      });
+    }
+  };
```
dolnaro, line 32: I didn't understand what this if statement does and couldn't test because of the npm test error. Can you help me to understand?

dolnaro (approved):

src/App.tsx
```diff
@@ -1,7 +1,47 @@
...
+  handleKeyEvent = (event: KeyboardEvent) => {
+    this.setState({
+      key: event.key,
+    });
+
+    if (event.key === ' ') {
+      this.setState({
+        key: 'Space',
+      });
+    }
```
meforoma, line 31: good, you went beyond - to actually spell the `Space`

@dolnaro if not this code, when pressing space on keyboard just an empty space will render - this code makes an explicit `Space` word to render

meforoma (approved):
👏🏼

PR #441, opened 2023-01-01
react_keyboard

DiDemkiv:
[DEMO LINK](https://DiDemkiv.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.recordKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.recordKey);
+  }
+
+  recordKey = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        {this.state.pressedKey
+          ? (
+            <p className="App__message">{`The last pressed key is [${this.state.pressedKey}]`}</p>
+          )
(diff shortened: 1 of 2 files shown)


batstolya:
everything is fine, but I suggest to optimize your code a bit)

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+    return (
+      <div className="App">
+        {this.state.pressedKey
+          ? (
+            <p className="App__message">{`The last pressed key is [${this.state.pressedKey}]`}</p>
+          )
+          : (
+            <p className="App__message">Nothing was pressed yet</p>
+          )}
+      </div>
```
batstolya, line 30: you can use destructuring, it's optional but I think it's better
```suggestion
  render() {
    return (
      <div className="App">
        {pressedKey
          ? (
            <p className="App__message">{`The last pressed key is [${pressedKey}]`}</p>
          )
          : (
            <p className="App__message">Nothing was pressed yet</p>
          )}
      </div>
```

batstolya, line 30: you can also simplify this code. Since our paragraphs are the same, we can render inside it.
```suggestion
    const { pressedKey } = this.state;

    return (
      <div className="App">
        <p className="App__message">
          {pressedKey
            ? `The last pressed key is [${pressedKey}]`
            : 'Nothing was pressed yet'}
        </p>
      </div>
    );
```

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: '',
+  };
```
batstolya, line 6: also you forgot about TypeScript)
```suggestion
  type State = {
  pressedKey: string | null;
};

export class App extends React.Component<{}, State> {
  state: Readonly<State> = {
    pressedKey: '',
  };
```

batstolya (approved):
good job!

valerii-derkach (approved):

PR #450, opened 2023-01-03
solution

kenvastolord:
[DEMO LINK](https://kenvastolord.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,41 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+type Props = {};
+
+export class App extends React.Component<Props, State> {
```
dolnaro, line 9: I don't think you should use `type Props` because you don't have a type definition here and it works if you remove type Props and add only `<{}, State>` to row 9.

Just to make it shorter with less code :)

dolnaro (approved):

meforoma (approved):
nice ✅

PR #451, opened 2023-01-03
keyboard

grslckr:
[DEMO LINK](https://grslckr.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,39 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type AppProps = {};
+
+type AppState = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<AppProps, AppState> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (e) => this.keyUpHandler(e));
+  }
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type AppProps = {};
+
+type AppState = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<AppProps, AppState> {
```
meforoma, line 9: empty object does not deserve the type 😉 
```suggestion
type AppState = {
  pressedKey: string;
};

export class App extends React.Component<{}, AppState> {
```

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (e) => this.keyUpHandler(e));
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.keyUpHandler);
+  }
```
meforoma, line 20: why adding and removing the listener in a different manner?
it's still a callback here and there - `this.keyUpHandler` is perfectly fine
```suggestion
  componentDidMount() {
    document.addEventListener('keyup', this.keyUpHandler);
  }

  componentWillUnmount() {
    document.removeEventListener('keyup', this.keyUpHandler);
  }
```

meforoma (approved):
✅

PR #453, opened 2023-01-03
React_KeyBoard was done

illia-kots:
[DEMO LINK](https://illia-kots.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,39 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | boolean,
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: false,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleDocumentKeyUp);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleDocumentKeyUp);
+  }
+
+  handleDocumentKeyUp = (event: KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | boolean,
```
polosanya, line 4: Is it really need to be string or boolean? Can't it be only string type?
You can check if it's an empty string, which means nothing was pressed

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: false,
+  };
+
+  componentDidMount(): void {
```
polosanya, line 12:
```suggestion
  componentDidMount() {
```

polosanya (changes requested):
Almost perfect! Please check the comments to improve your code a bit)

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {pressedKey.length === 0
+            ? 'Nothing was pressed yet'
+            : `The last pressed key is [${pressedKey}]`}
```
Anastasiia-Svintsova, line 34:
```suggestion
          {!pressedKey.length
            ? 'Nothing was pressed yet'
            : `The last pressed key is [${pressedKey}]`}
```

Anastasiia-Svintsova (approved):

src/App.tsx
illia-kots: Ok. Thanks!

PR #455, opened 2023-01-03
Solution

UmizDemud:


src/App.tsx
```diff
@@ -1,7 +1,30 @@
...
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    lastKey: '',
+    listener: (event: KeyboardEvent) => {
+      this.setState({ lastKey: event.key });
+    },
+  };
+
```
meforoma, line 10: not a good idea to put the callback inside the state - it's not the state by nature
```suggestion
  state = {
    lastKey: '',
  };
  
  listener = (event: KeyboardEvent) => {
    this.setState({ lastKey: event.key });
  }
  // i think you'll just need to declare it below lifecycle methods
  // and then you just use it like `this.listener` on adding and removing listeners

```

src/App.tsx
```diff
@@ -1,7 +1,30 @@
...
+  render() {
+    return (
+      <div
+        className="App"
+      >
+        {this.state.lastKey
+          ? <p className="App__message">{`The last pressed key is [${this.state.lastKey}]`}</p>
+          : <p className="App__message">Nothing was pressed yet</p>}
+      </div>
```
meforoma, line 27: you're rendering `<p>` anyway right?
so why not just conditionally set the contents inside <p> instead of repeating the <p> twice?
```suggestion
      <p className="App__message">
        {this.state.lastKey
          ? `The last pressed key is [${this.state.lastKey}]`
          : 'Nothing was pressed yet'}
        </p>
      </div>
```

src/App.tsx
```diff
@@ -1,7 +1,30 @@
...
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.state.listener);
+  }
+
+  render() {
+    return (
+      <div
+        className="App"
+      >
```
meforoma, line 23: no need for multi-line here
```suggestion
      <div className="App">
```

meforoma (approved):
✅

PR #456, opened 2023-01-03
Solution

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
+            this.state.value
+              ? `The last pressed key is [${this.state.value}]`
+              : 'Nothing was pressed yet'
+          }
```
meforoma, line 32: the only thing we've missed is making usage of `this.state.value` shorter:
```
render () {
  const { value } = this.state;
  ...
```

meforoma (approved):
👍🏼

PR #470, opened 2023-01-17
Develop

katya-suvorova:
https://katya-suvorova.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,27 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+type State = {
+  message: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    message: 'Nothing was pressed yet',
+  };
+
+  handleClick = (event: KeyboardEvent) => {
```
polosanya, line 12: Let's fix this problem
image

polosanya (changes requested):
Almost done!

qoa11a (approved):

PR #472, opened 2023-01-19
solution

Malva37:
[DEMO LINK](https://Malva37.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey:string,
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.onKeyDown);
+  }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.onKeyDown);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('keyup', this.onKeyDown);
```
nzrii-kysel, line 17:
```suggestion
    document.removeEventListener('keyup', this.onKeyDown);
```

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <p className="App__message">
+        {pressedKey
+          ? `The last pressed key is [${pressedKey}]`
+          : 'Nothing was pressed yet'}
+      </p>
+    );
```
nzrii-kysel, line 33: you are using BEM element without the block

nzrii-kysel (approved):

src/App.tsx
Denchepornyuk: should be `document` here

Denchepornyuk (changes requested):

denys-cheporniuk (approved):

PR #475, opened 2023-01-20
solution

divineretribution1:
[DEMO LINK](https://divineretribution1.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.pressedKeyValue);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.pressedKeyValue);
+  }
+
+  pressedKeyValue = (event: KeyboardEvent) => {
```
agniya-i, line 20: not a big mistake, but it would be better for function name specify the action, for example, `pressKey`, `handlePressKey` or  `handleKeyup` etc.

agniya-i (approved):

PR #479, opened 2023-01-23
React Keyboard - completed

GarryRocksman:
[DEMO](https://GarryRocksman.github.io/react_keyboard/)

 - Have a nice Day 🔥

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,41 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: null,
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.keyupHandler);
+  }
+
+  componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: null,
+  };
```
nzrii-kysel, line 6: create type for state

nzrii-kysel (changes requested):

src/App.tsx
GarryRocksman: Thanks, fixed

alexandra-protyanova (approved):
🚀

src/App.tsx
```diff
@@ -1,7 +1,44 @@
...
+      pressedKey: event.key,
+    });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        {this.state.pressedKey
+          ? (
+            <p className="App__message">
```
alexandra-protyanova, line 30: I would render `p` tag before ternary, to prevent duplicating

PR #482, opened 2023-01-23
Solution

RomeoKs:
[DEMO LINK](https://RomeoKs.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -17,15 +17,11 @@ export const App: React.FC = () => {
     <div className="App">
       {/* Keep this input for tests */}
       <input type="text" className="hidden" />
-      {keyPressed
-        ? (
-          <p className="App__message">
-            The last pressed key is [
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+      <input type="text" className="hidden" />
+      {keyPressed
+        ? (
+          <p className="App__message">
+            The last pressed key is [
+            {keyPressed}
+            ]
+          </p>
+        )
+        : <p className="App__message">Nothing was pressed yet</p>}
```
naziamoff, line 28: let's do it inside one `<p>`, I see no reason to conditionally render different tags here

naziamoff (approved):
good job

src/App.tsx
RomeoKs: Done.
Thank you!

PR #490, opened 2023-01-24
Solution by DK

DimaK123:


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+    key: null,
+  };
+
+  message = 'Nothing was pressed yet';
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ key: event.key });
+    });
+  }
```
polosanya, line 18: image
Also, I think it's better to save the function in some method

polosanya (changes requested):
Let's improve your code a bit

polosanya (approved):

PR #491, opened 2023-01-25
solution

Oleksandr-Maly:
[DEMO LINK](https://oleksandr-maly.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,11 +26,13 @@ export class App extends React.Component {
 
     return (
       <div className="App">
-        {
-          !(key)
-            ? <p className="App__message">Nothing was pressed yet</p>
-            : <p className="App__message">{`The last pressed key is [${key}]`}</p>
-        }
+        <p className="App__message">
+          {
+            !(key)
+              ? 'Nothing was pressed yet'
+              : `The last pressed key is [${key}]`
+          }
+        </p>
       </div>
     );
   }


src/App.tsx
```diff
@@ -1,7 +1,25 @@
...
+    ));
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">{this.state.message}</p>
+      </div>
+    );
+  }
```
oleh-holovnykh, line 24: 💣 state

```suggestion
  render() {
  
    const { message } = this.state;
  
    return (
      <div className="App">
        <p className="App__message">{ message }</p>
      </div>
    );
  }
```

src/App.tsx
```diff
@@ -1,7 +1,25 @@
...
+export class App extends React.Component {
+  state: Readonly<State> = {
+    message: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => (
+      this.setState({ message: `The last pressed key is [${event.key}]` })
+    ));
+  }
```
oleh-holovnykh, line 16: i think its good to add separate method for this function

```suggestion
 handleKeyUp = (event: KeyboardEvent) => (
      this.setState({ message: `The last pressed key is [${event.key}]` })
    );

  componentDidMount() {
    document.addEventListener('keyup', this.handleKeyUp);
  }
```

src/App.tsx
```diff
@@ -1,7 +1,25 @@
...
+
```
oleh-holovnykh, line 17: Add componentWillUnmount() to remove eventListener, for the case if we remove this component from page

oleh-holovnykh (changes requested):
Good job! 🔥👍 Cool solution with message in state

src/App.tsx
```diff
@@ -1,7 +1,33 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string;
```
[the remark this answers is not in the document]
Oleksandr-Maly: why it couldn't be a message?

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  render() {
+    const { key } = this.state;
+
+    return (
+      <div className="App">
+        {
+          !(key)
+            ? <p className="App__message">Nothing was pressed yet</p>
+            : <p className="App__message">{`The last pressed key is [${key}]`}</p>
+        }
```
Anna-Tkachenko, line 33: you repeated the code with adding paragraph two times
let's move the check inside of the paragraph, and will render corresponding message according to the `key` variable

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #509, opened 2023-01-27
solution

AndreyBerezyanskiy:
[DEMO LINK](https://AndreyBerezyanskiy.github.io/react_keyboard/)


src/App.tsx
```diff
@@ -1,7 +1,50 @@
...
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  isPressedKey: boolean,
+  currentKey: string,
+};
```
tania-boiar, line 6: You could use here just one variable.

src/App.tsx
```diff
@@ -1,7 +1,50 @@
...
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    isPressedKey: false,
+    currentKey: '',
+  };
```
tania-boiar, line 12: Here the same. Just `currentKey: '',` would be enough. It's already a falsy value.

src/App.tsx
```diff
@@ -1,7 +1,50 @@
...
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handlePress);
+  }
+
+  handlePress = (event: KeyboardEvent) => {
+    this.setState({
+      isPressedKey: true,
+      currentKey: event.key,
+    });
+  };
```
tania-boiar, line 27: And here could use only `currentKey: event.key`

src/App.tsx
```diff
@@ -1,7 +1,50 @@
...
+
+  render() {
+    const { isPressedKey, currentKey } = this.state;
+
+    return (
+      <div className="App">
+        {isPressedKey
```
tania-boiar, line 34: And also here check if `currentKey,`

tania-boiar (approved):
Good job! 😉
Interesting solution.

Left you a couple of comments. You could look through them if you want to.

src/App.tsx
AndreyBerezyanskiy: thanks, agree with you

src/App.tsx
AndreyBerezyanskiy: useful advice! thank you)

src/App.tsx
```diff
@@ -1,7 +1,47 @@
...
+              The last pressed key is [
+              {currentKey}
+              ]
+            </p>
+          )
+          : (
+            <p className="App__message">
+              Nothing was pressed yet
+            </p>
+          )}
```
ArturErmolenko, line 43:
```suggestion
     <p className="App__message">
        {currentKey
          ? (`The last pressed key is [${currentKey ]`)
          : ( ' Nothing was pressed yet' )}
            </p>
```
Follow DRY

ArturErmolenko (approved):

PR #515, opened 2023-01-28
keybord

Kukharuk:
[DEMO LINK](https://Kukharuk.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,12 @@
 import { Component } from 'react';
 
 type State = {
-  previous: string,
+  pressedButton: string,
 };
 
 export class App extends Component<{}, State> {
   state = {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  previous: string,
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    previous: '',
```
qoa11a, line 9: why is it called previous? what does it represent? it should be named "key" or something like that

qoa11a (changes requested):

mykhalenych (approved):

PR #516, opened 2023-01-28
solution

metalikaqq:
[DEMO LINK](https://metalikaqq.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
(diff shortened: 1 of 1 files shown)


svitjojo (changes requested):
The code is good 👍 
But your demo link is not working, check it.

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  handleKey = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.pressedKey
```
Anna-Tkachenko, line 28: will be better to get needed properties from the state via desctructuring before the `return` statement

Anna-Tkachenko (approved):

PR #519, opened 2023-01-29
Solution

OleksandrOse:
[DEMO LINK](https://OleksandrOse.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  text: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    text: 'Nothing was pressed yet',
+  };
+
+  UNSAFE_componentWillMount() {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    text: 'Nothing was pressed yet',
+  };
+
+  UNSAFE_componentWillMount() {
+    document.removeEventListener('keyup', this.handleClickButton);
+  }
```
SanyaBratashchuk, line 14: Why do you need this method? Why can't you just use componentDidMount?

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  text: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    text: 'Nothing was pressed yet',
```
SanyaBratashchuk, line 9: set empty string or null as default value

SanyaBratashchuk (changes requested):

tetiana-ivanchenko (approved):

PR #520, opened 2023-01-29
Develop

Kanzatuwka:
[DEMO LINK](https://kanzatuwka.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import { Component } from 'react';
+import './index.scss';
 
 type State = {
   pressedKey: string;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.updatePressedKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.updatePressedKey);
+  }
+
+  updatePressedKey = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,32 @@
...
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
```
Volodymyr-Melnyk-Fron10der, line 11: 1. State has to be created in constructor
2. String `'Nothing was pressed yet'` is not the key type, so I suggest to set `pressedKey: null` by default and then make condition in your JSX
```suggestion
  constructor(props) {
    super(props);
    this.state = { pressedKey: null },
  }
```

src/App.tsx
```diff
@@ -1,7 +1,32 @@
...
+  componentDidMount() {
+    document.addEventListener('keyup', this.updatePressedKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.updatePressedKey);
+  }
+
+  updatePressedKey = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
```
Volodymyr-Melnyk-Fron10der, line 22: String `The last pressed key is [${event.key}]` is not a key type
```suggestion
    this.setState({ pressedKey: event.key });
```

src/App.tsx
```diff
@@ -1,7 +1,32 @@
...
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">{this.state.pressedKey}</p>
```
Volodymyr-Melnyk-Fron10der, line 28:
```suggestion
        <p className="App__message">{
          !!this.state.pressedKey
            ? `The last pressed key is [${this.state.pressedKey}]`
            : 'Nothing was pressed yet'
        }</p>
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, good job! But We need to be consistance with 'Single responsibility' pattern. If you named key as 'pressedKey' then there should be ONLY the key value and nothing else

Kanzatuwka:
Since the constructors is not popular at the moment, and the conditions for the task do not mention its use, I did not use it. All other suggestions for improving code  have been taken into account.

tetiana-ivanchenko (approved):

PR #522, opened 2023-01-29
solution react_keyboard

andrii-kurash:
Have a great day!🔥

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,25 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string;
+};
+
+export class App extends React.Component {
+  state: Readonly<State> = {
+    message: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => (
+      this.setState({ message: `The last pressed key is [${event.key}]` })
+    ));
+  }
+
+  render() {
+    return (
+      <div className="App">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,25 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string;
+};
+
+export class App extends React.Component {
+  state: Readonly<State> = {
+    message: 'Nothing was pressed yet',
```
Anna-Tkachenko, line 9: in state you need to store pressed key, not message
show the message if nothing is pressed directly in render method

src/App.tsx
```diff
@@ -1,7 +1,25 @@
...
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => (
+      this.setState({ message: `The last pressed key is [${event.key}]` })
+    ));
+  }
+
```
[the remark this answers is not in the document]
Anna-Tkachenko: fix @anna-inozemets comment

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string;
```
Anna-Tkachenko, line 4: it's not the message, you need to store pressed key

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #524, opened 2023-01-29
completed

AlexPro100pro:


AlexPro100pro:
[DEMO LINK](https://AlexPro100pro.github.io/react_keyboard/)

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
-  </div>
-);
+type State = {
+  pressedKey: string,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
```
Anastasiia-Khudych, line 10: It's not a `pressedKey`, it's the message you show on the page. I suppose it's better to store in the state `pressedKey ` value (default = ' ' or null) and render the page according to its state.

Anastasiia-Khudych (changes requested):

src/App.tsx
Anna-Tkachenko: fix @Anastasiia-Khudych comment

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #531, opened 2023-01-30
Solution

alSemenyuk:
https://mate-academy.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string
+};
+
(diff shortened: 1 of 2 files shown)


nkondd (approved):
Good job 🔥 
Unusual and interesting solution with `state.message` 🤔

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  message: string
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    message: 'Nothing was pressed yet',
```
Anna-Tkachenko, line 9: you need to store pressed key in state, not the message

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #532, opened 2023-01-30
Develop

Sergey-Mironenko:


Sergey-Mironenko:
https://Sergey-Mironenko.github.io/react_keyboard/

tetiana-ivanchenko (changes requested):
please, check that all point from task are done, before creating PR
image

src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keydown', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
```
tetiana-ivanchenko, line 16: if you use the same code more than once - it's make sense to create serapate variable for it
move callback in the separate variable

src/App.tsx
```diff
@@ -1,7 +1,36 @@
+//  import { prependOnceListener } from 'process';
```
tetiana-ivanchenko, line 1: remove comments

tetiana-ivanchenko (changes requested):

AlisaVlasova (approved):

PR #539, opened 2023-02-01
react keyboard v0.1

fenderukr:
https://fenderukr.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,28 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  // eslint-disable-next-line react/sort-comp
+  handleDocumentClick = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleDocumentClick);
+  }
+
+  componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,28 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
agniya-i, line 5: `pressedKey` looks like page message, not actually a pressed key. It's would be better to store in the state `pressedKey` value (as default you can store `' ' `or `null` ) and render the page message according to this state

src/App.tsx
```diff
@@ -1,7 +1,28 @@
...
+  };
+
+  // eslint-disable-next-line react/sort-comp
+  handleDocumentClick = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
```
agniya-i, line 10:
```suggestion
    this.setState({ pressedKey: event.key });
```

src/App.tsx
fenderukr: thanks, fixed

src/App.tsx
fenderukr: fixed

src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+      );
+    }
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          The last pressed key is
+          [
+          {this.state.pressedKey}
+          ]
```
witflash, line 36: You can use string interpolation to avoid multiline syntax.

witflash (approved):
Well done 👍

PR #540, opened 2023-02-01
Solution

illia-kyselov:
[DEMO LINK](https://illia-kyselov.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  phrase: string,
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    phrase: '',
+  };
+
+  componentDidMount() {
+    this.setState({ phrase: 'Nothing was pressed yet' });
```
TarasHoliuk, line 13: instead of changing state here, set `'Nothing was pressed yet'` by default

TarasHoliuk (changes requested):

nazarmatsevych (approved):

PR #550, opened 2023-02-07
React develop

Vasyliy22:
[DEMO LINK](https://Vasyliy22.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,7 +4,7 @@ type State = {
   pressedKey: string;
 };
 
-export class App extends Component<State> {
+export class App extends Component<{}, State> {
   state = {
     pressedKey: 'Nothing was pressed yet',
   };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handlepressedKey);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('keyup', this.handlepressedKey);
+  }
+
+  handlepressedKey = (event: KeyboardEvent) => {
```
agniya-i, line 20:
```suggestion
  handlePressKey = (event: KeyboardEvent) => {
```

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
agniya-i, line 9: it looks like the message, not a pressed key, so it would be better to set just event key for `pressedKey`   and display the message depending on what is in `pressedKey`

agniya-i (changes requested):

qoa11a (approved):

PR #556, opened 2023-02-12
React_keyboard

07Spirit:
[DEMO LINK](https://07Spirit.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,41 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  myKey: string | null,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    myKey: null,
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.lastKeyClick);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.lastKeyClick);
+  }
+
+  lastKeyClick = (elem: KeyboardEvent) => {
+    this.setState({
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+
+  render(): React.ReactNode {
+    return (
+      <div className="App">
+        {this.state.myKey
+          ? (
+            <p className="App__message">
+              The last pressed key is [
+              {this.state.myKey}
+              ]
```
witflash, line 34: Better use string interpolation instead of multiplying strings.

src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+            </p>
+          )
+          : (<p className="App__message">Nothing was pressed yet</p>)}
```
witflash, line 37: If you write an entire element in a single string you can avoid wrapping it with parentheses:
```
: <p className="...">...</p>
```

src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.lastKeyClick);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.lastKeyClick);
+  }
+
+  lastKeyClick = (elem: KeyboardEvent) => {
```
witflash, line 20: It's better to name it smth like `handleKeyClick`

witflash (changes requested):
Good work.
Check my comments, please. And try to improve your code.

src/App.tsx
07Spirit: Okey

src/App.tsx
07Spirit: Ok

nazarmatsevych (approved):

PR #558, opened 2023-02-13
done

AntonChetverykMA:
[DEMO LINK](https://AntonChetverykMA.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.onClick);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.onClick);
+  }
+
+  onClick = (e: KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+      pressedKey: e.key,
+    });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">{`The last pressed key is [${pressedKey}]`}</p>
```
tetiana-ivanchenko, line 27: if nothing was pressed there should be no text "The last pressed key is"

tetiana-ivanchenko (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
tetiana-ivanchenko, line 5: actually, the name must match the content, so rename it please

tetiana-ivanchenko (changes requested):

src/App.tsx
AntonChetverykMA: @tetiana-ivanchenko 
Снимок экрана (56)

AntonChetverykMA: but i can change it, no proplem)

sTorba24 (approved):
Good work!

PR #565, opened 2023-02-20
react keyboard

Mykola-Busarhin:
[DEMO LINK](https://Mykola-Busarhin.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.showLetter);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.showLetter);
+  }
+
+  showLetter = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is ${event.key}` });
+  };
+
+  render() {
+    return (
+      <div className="App">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
Volodymyr-Melnyk-Fron10der, line 9: Keep single responsibility pattern, if it's the `pressedKey` then in this variable should be value of pressed key only, no other text, so set it `''` by default

Volodymyr-Melnyk-Fron10der (changes requested):
Looks good👍 just one remark

src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+  showLetter = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is ${event.key}` });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        {this.state.pressedKey === ''
+          ? (
+            <p className="App__message"> Nothing was pressed yet </p>)
```
TarasHoliuk, line 29:
```suggestion
            <p className="App__message">Nothing was pressed yet</p>)
```
redundant spaces

src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+  }
+
+  showLetter = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is ${event.key}` });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        {this.state.pressedKey === ''
```
TarasHoliuk, line 27:
```suggestion
        {this.state.pressedKey
```
you may check it in this way - an empty string will give you false

FYI

TarasHoliuk (approved):
Approved :heavy_check_mark: 

Added 2 comments FYI (no need to re-request review)

PR #569, opened 2023-02-24
Solution

katerinalex:
[DEMO LINK](https://katerinalex.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,46 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+// export const App: React.FC = () => (
+//   <div className="App">
+//     <p className="App__message">The last pressed key is [Enter]</p>
+//   </div>
+// );
+export class App extends React.Component {
+  state = {
+    key: 'Nothing',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleDocumentKey);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+// export const App: React.FC = () => (
+//   <div className="App">
+//     <p className="App__message">The last pressed key is [Enter]</p>
+//   </div>
+// );
```
qoa11a, line 7: remove comments

src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
+export class App extends React.Component {
+  state = {
+    key: 'Nothing',
```
qoa11a, line 10: you should store null here if no key was pressed

src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
+        </div>
+      );
+    }
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          The last pressed key is
+          {' '}
+          {this.state.key}
```
qoa11a, line 41: then here you can use ternary operator to show either that no key was pressed or the key that was pressed

qoa11a, line 41: this should be wrapped inside "[" "]"

qoa11a (changes requested):

agniya-i (approved):

PR #574, opened 2023-02-27
solution

bohdan-gavrilyk:
https://bohdan-gavrilyk.github.io/react_keyboard/


slavanosatskyi (approved):
Good work 👍 Only one remark - you should always try to give meaningful variables names.. str is not the best choice. Try to come up with a meaning which should variable convey. In this case it can be message or something like that

PR #575, opened 2023-02-27
solution

vlad-trundenko:
[DEMO LINK](https://vlad-trundenko.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    key: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleClick);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleClick);
+  }
+
+  handleClick = (event: KeyboardEvent) => {
+    this.setState({ key: `The last pressed key is  [${event.key}]` });
+  };
+
+  render() {
+    return (
+      <div className="App">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    key: 'Nothing was pressed yet',
```
qoa11a, line 9: it's not a key. It should be either null or pressed key

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleClick);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleClick);
+  }
+
+  handleClick = (event: KeyboardEvent) => {
+    this.setState({ key: `The last pressed key is [${event.key}]` });
```
qoa11a, line 21: here as well. Just set it to key

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">{this.state.key}</p>
```
qoa11a, line 27: and here you can have
{this.state.key ? \`The last pressed key is [${this.state.key}]\` : 'Nothing was pressed yet'}

qoa11a (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,35 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string | null
```
polosanya, line 4:
```suggestion
  pressedKey: string | null
```
According to the task description, it should be called like this
image

polosanya (approved):
Well done!

PR #587, opened 2023-03-06
Develop

Roman-Malysh:
https://Roman-Malysh.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+class App extends React.Component {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,29 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
tetiana-ivanchenko, line 5: not so good approach for naming, it can be two ways: if you left as it is, you should name variable, smth like "pressedKeyMessage", or here you should save only pressed key

tetiana-ivanchenko (approved):

PR #591, opened 2023-03-07
Solution

stanislavcodes:
[DEMO LINK](https://stanislavcodes.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,7 @@
 import { Component } from 'react';
 
 type State = {
-  pressedKey: string | '';
+  pressedKey: string;
 };
 
 export class App extends Component<{}, State> {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | null;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.keyPressHandler);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.keyPressHandler);
+  }
+
+  keyPressHandler = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | null;
```
kirill-vypirovskyi, line 4: null is reduntant

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: null,
```
kirill-vypirovskyi, line 9: You can just empty string as default value, it is falsy too and will work

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {pressedKey
+            ? `The last pressed key is [${pressedKey}]`
+            : 'Nothing was pressed yet'}
```
kirill-vypirovskyi, line 32: I'm not sure

```suggestion
          {
            pressedKey
              ? `The last pressed key is [${pressedKey}]`
              : 'Nothing was pressed yet'
           }
```

kirill-vypirovskyi (approved):

src/App.tsx
OleksiiBlezniuk: Agree, let's use falsy values of same type to set initial state value, empty string will be good here

src/App.tsx
OleksiiBlezniuk: @kirill-vypirovskyi you are right, this format is more readable

OleksiiBlezniuk (changes requested):
Let's use falsy values of same type for state initial values

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | '';
+};
```
OleksiiBlezniuk, line 5:
```suggestion
type State = {
  pressedKey: string;
};
```

OleksiiBlezniuk: Empty string has `string` type too

OleksiiBlezniuk (approved):

PR #593, opened 2023-03-07
Solution

Vlad-Fedorishchev:
Demo Link  -  https://Vlad-Fedorishchev.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.eventHandler);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.eventHandler);
+  }
+
+  eventHandler = (event: KeyboardEvent) => {
+    const { key, code } = event;
+
+    const isSpaceKey = code === 'Space'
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.eventHandler);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.eventHandler);
+  }
+
+  eventHandler = (event: KeyboardEvent) => {
```
polosanya, line 16: image
```suggestion
  handleClick = (event: KeyboardEvent) => {
```

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    const { key, code } = event;
+
+    const isSpaceKey = code === 'Space'
```
polosanya, line 19: Why is it named like a boolean value?

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+      ? 'The last pressed key is [Space]'
+      : `The last pressed key is [${key}]`;
```
polosanya, line 21: No need to set the whole sentence as a pressed key value. You should store only the key as a pressedKey value

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    this.setState({ pressedKey: isSpaceKey });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {pressedKey}
```
polosanya, line 32: You can make pressedKey null by default and render the whole sentence or "Nothing was pressed" here using ternary operator

polosanya (changes requested):
Let's improve your code a bit)

agniya-i (approved):

PR #603, opened 2023-03-12
initial solution

Maks-Tsarenko:
[DEMO LINK](https://Maks-Tsarenko.github.io/react_keyboard/)


OlEzhA0 (approved):
Hey! I approve your work, but would be better to rewrite it with function component

PR #606, opened 2023-03-13
Solution

Ihor-Zakharov:
[DEMO LINK](https://ihor-zakharov.github.io/react_keyboard/)


DarkMistyRoom (changes requested):
use `removeEventListener `to remove a global handler in `componentWillUnmount`.

PR #618, opened 2023-03-15
solution

Nordexz:
[DEMO LINK](https://nordexz.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
+import { Keyboard } from './keyboard';
 
 export const App: React.FC = () => (
   <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
+    <p className="App__message"><Keyboard /></p>
   </div>
 );
--- a/src/keyboard.tsx
+++ b/src/keyboard.tsx
@@ -1,0 +1,33 @@
+import { Component, ReactNode } from 'react';
+
+type State = {
+  key: string,
+};
+
+export class Keyboard extends Component<{}, State> {
+  state: Readonly<State> = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.keySaver);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.keySaver);
+  }
+
+  keySaver = (event: KeyboardEvent) => {
+    this.setState({ key: event.key });
+  };
+
+  render(): ReactNode {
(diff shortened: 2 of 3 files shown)


src/keyboard.tsx
```diff
@@ -0,0 +1,33 @@
+import { Component, ReactNode } from 'react';
+
+type State = {
+  key: string,
```
NikitaStambul, line 4: "Make the `App` a class component with `pressedKey` in the `state`." -requirement  from readme.md file

src/keyboard.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.keySaver);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.keySaver);
+  }
+
+  keySaver = (event: KeyboardEvent) => {
```
NikitaStambul, line 20: function name should be a verb, not a noun
it will be better to name it like an action that it does
'saveKey' or 'handlePressedKey' and so on

src/App.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
+import { Keyboard } from './keyboard';
 
 export const App: React.FC = () => (
   <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
+    <p className="App__message"><Keyboard /></p>
```
NikitaStambul, line 6: "Make the App a class component with pressedKey in the state." you don't need separate Keyboard component but make an App class component to do your logic

NikitaStambul (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.saveKey);
+  }
+
+  saveKey = (event: KeyboardEvent) => {
+    this.setState({ key: event.key });
+  };
+
+  render(): ReactNode {
```
illiarerun, line 24: overkill

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+    const { key } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
+            key
+              ? `The last pressed key is [${key}]`
+              : 'Nothing was pressed yet'
+          }
```
illiarerun, line 34: you have no jsx here, better store it in variable to improve readability

PR #621, opened 2023-03-15
Keyboard

Igor-Yaremchuk:
[Keyboard](https://gor-Yaremchuk.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleClick);
+  }
+
+  componentWillUnmount() {
+    setTimeout(() => {
+      document.removeEventListener('keyup', this.handleClick);
+    }, 3000);
+  }
+
+  handleClick = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last prerssed key is [${event.key}]` });
+  };
+
+  render() {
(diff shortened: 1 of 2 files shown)


mykhalenych (changes requested):
your deploy link isn't working

Igor-Yaremchuk:
> your deploy link isn't working

Fixed

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
witflash, line 9: Your state is about the pressed key. So if there is no pressed key it should be `null` instead of some info message.
And the message you can render while you check the pressed key (see next comment)

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  }
+
+  componentWillUnmount() {
+    setTimeout(() => {
+      document.removeEventListener('keyup', this.handleClick);
+    }, 3000);
+  }
+
+  handleClick = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last prerssed key is [${event.key}]` });
```
witflash, line 23: Here you should set just a key name:
```
this.setState({ pressedKey: event.key })
```

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p>
+          {pressedKey}
```
witflash, line 32: And here you can check if any key has been pressed and render a required message:
```
{pressedKey ? `The last pressed key is [${pressedKey}]` : 'Nothing was pressed yet'}
```

witflash (changes requested):
Good job!
Check my comments, please. And let's do some improvements.

src/App.tsx
Igor-Yaremchuk: Thanks)

sTorba24 (approved):
Cool!

PR #628, opened 2023-03-15
React keyboard

andriibezkorovainyi:
[DEMO LINK](https://andriibezkorovainyi.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,41 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | null
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: null,
```
Yekatery, line 9: why did you set null here instead of string?

Yekatery (approved):

PR #636, opened 2023-03-15
solution

valeriy-mikheyev:
[DEMO LINK](https://valeriy-mikheyev.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,12 @@
 import React from 'react';
 
 interface State {
-  pressedKey: string | null;
+  pressedKey: string;
 }
 
 export class App extends React.Component<{}, State> {
   state: State = {
-    pressedKey: null,
+    pressedKey: '',
   };
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,42 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  pressedKey: string | null;
```
vlad-demchuk, line 4: you could use just an empty string instead of `null`, just FYI, not a mistake

src/App.tsx
```diff
@@ -1,7 +1,42 @@
...
+    const { pressedKey } = this.state;
+
+    return (
+      <div>
+        {pressedKey ? (
+          <p>
+            The last pressed key is
+            [
+            {pressedKey}
+            ]
```
vlad-demchuk, line 34: it will be clearer to use interpolation here

vlad-demchuk (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  handleKeyUp = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div>
```
vlad-demchuk, line 28: also, would be good to fix the tests
image

hovewer, I didn't mention it in the previous review, so let it go)

vlad-demchuk (approved):

PR #664, opened 2023-03-29
react_keyboard

MaksymHuriev:
[DEMO LINK](https://MaksymHuriev.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  lastPressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    lastPressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener(
+      'keydown', this.keyboardHandler,
+    );
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener(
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  lastPressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    lastPressedKey: 'Nothing was pressed yet',
```
Volodymyr-Melnyk-Fron10der, line 9: 'Nothing was pressed yet' - is not a key, let's keep single responsibility pattern, so lastPressedKey is used only for saving a key which was pressed

Volodymyr-Melnyk-Fron10der: set it '' by default and  just check in your JSX is some key pressed if no - show the default message

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener(
+      'keydown', this.keyboardHandler,
+    );
+  }
+
+  keyboardHandler = (event: KeyboardEvent) => {
+    this.setState({ lastPressedKey: `The last pressed key is [${event.key}]` });
```
Volodymyr-Melnyk-Fron10der, line 25:
```suggestion
    this.setState({ lastPressedKey: event.key });
```

Volodymyr-Melnyk-Fron10der (changes requested):
Good job) check my remarks plz

SerhiiKirik (approved):
Great 🔥

PR #671, opened 2023-04-03
solution

anna-akhalmosulishvili:
https://anna-akhalmosulishvili.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          { pressedKey.length !== 0
```
SerhiiKirik, line 26: You can make it like that
```suggestion
          {pressedKey.length
```

SerhiiKirik (approved):
Great 🔥

PR #678, opened 2023-04-06
created a keyboard clicker

prokivskaaa:
[DEMO LINK](https://prokivskaaa.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  pressedKey: string | null;
+}
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  pressedKey: string | null;
+}
+
+export class App extends React.Component {
+  state: Readonly<State> = {
+    pressedKey: null,
```
Esteban4769, line 9: Not sure this is right way to set default value for string.

Esteban4769 (approved):
Good job!

src/App.tsx
jstmpelowycz: that's right

jstmpelowycz (approved):

PR #685, opened 2023-04-10
added class component that display last pressed button

Esteban4769:


src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+
+  handleKeyPress = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+    const message = pressedKey.length > 0
+      ? `The last pressed key is [${pressedKey}]`
+      : 'Nothing was pressed yet';
```
PavloReutskiy, line 28:
```suggestion
    const message = pressedKey
      ? `The last pressed key is [${pressedKey}]`
      : 'Nothing was pressed yet';
```

start value of pressedKey = '' - it's 'false', so you don't need to check the length, it will work ok

PavloReutskiy (approved):
looks good! 👍

src/App.tsx
jstmpelowycz: agree, let's fix

src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
```
jstmpelowycz, line 5: let's make it nullable instead, more consistent
```suggestion
type State = {
  pressedKey: string | null;
};
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #687, opened 2023-04-10
made keyboard hendler

PavloReutskiy:
[DEMO LINK](https://PavloReutskiy.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,39 @@
-import React from 'react';
+import React, { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: '',
+  };
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
-import React from 'react';
+import React, { Component } from 'react';
```
igorkruz, line 1: `import { Component } from 'react';`

igorkruz (approved):
LGTM 👍

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
```
jstmpelowycz, line 4: let's assign null if nothing pressed yet
```suggestion
  pressedKey: string | null;
```

jstmpelowycz (changes requested):

src/App.tsx
PavloReutskiy: done!

jstmpelowycz (approved):

PR #688, opened 2023-04-11
created react keyboard

igorkruz:
[DEMO LINK](https://igorkruz.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.keyPressHandler);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.keyPressHandler);
+  }
+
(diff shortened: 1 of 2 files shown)


Pozdnya (approved):
Good job!!!

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: 'Nothing was pressed yet',
```
```suggestion
    pressedKey: null,
```

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+  };
+
+  componentDidMount(): void {
```
SerhiiKirik, line 12: It's extra, bc TypeScript already know about React types
```suggestion
  componentDidMount() {
```

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+    document.addEventListener('keyup', this.keyPressHandler);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.keyPressHandler);
+  }
+
+  keyPressHandler = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
```
SerhiiKirik, line 21: So, each time you set this string?
So, why do you give the name `pressedKey` for this string, it can be `Nothing was pressed yet`?
Just put a key into the state

SerhiiKirik (changes requested):
Great 🚀 

src/App.tsx
igorkruz: fixed

SerhiiKirik (approved):
LGTM 🔥

PR #689, opened 2023-04-11
add solution

Pozdnya:
[DEMO LINK](https://pozdnya.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: null | string,
+}
+
+export class App extends React.PureComponent<{}, State> {
+  state = {
+    key: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      const { key } = event;
+
+      return this.setState({ key });
+    });
+  }
+
+  render() {
+    const { key } = this.state;
+
+    return (
+      <div className="App">
+        {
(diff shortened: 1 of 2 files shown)


PashaMalyshkin (approved):
GJ Kyryl !!!

src/App.tsx
```diff
@@ -1,7 +1,39 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: null | string,
```
jstmpelowycz, line 4:
```suggestion
  key: string | null;
```

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+    const { key } = this.state;
+
+    return (
+      <div className="App">
+        {
+          !key
+            ? <p className="App__message">Nothing was pressed yet</p>
+            : <p className="App__message">{`The last pressed key is [${key}]`}</p>
+        }
+      </div>
```
jstmpelowycz, line 36:
```suggestion
      <div className="App">
        <p className="App__message">
          {
            key
              ? {`The last pressed key is [${key}]`}
              : Nothing was pressed yet
          }
        </p>
      </div>
```

jstmpelowycz (changes requested):

src/App.tsx
Pozdnya: Done.
But I should wrap it on single quotes, BCS React thing I use it like a variable.

src/App.tsx
Pozdnya: Done

jstmpelowycz (approved):

PR #693, opened 2023-04-12
Solution react_keyboard

NazariiAlieksieiev:
[DEMO LINK](https://NazariiAlieksieiev.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+  key: string
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', ({ key }) => {
```
etojeDenys, line 13: It's better to remove event listeners when the component is unmounted. Create a separate method for the event listener and use `componentWillUnmount` to remove it.

src/App.tsx
```diff
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
```
etojeDenys, line 1: Consider importing React as well, since it's required for JSX. Add `import React from 'react';` at the beginning of the file.

etojeDenys (changes requested):

PR #698, opened 2023-04-12
keyboard event

ofenchuk:
[DEMO LINK](https://ofenchuk.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,10 @@
-import React, { Component } from 'react';
+import React from 'react';
 
 type State = {
   key: string;
 };
 
-export class App extends Component<{}, State> {
+export class App extends React.Component<{}, State> {
   state = {
     key: '',
   };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import React, { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (e) => {
+      this.setState({ key: e.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.setState, false);
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.key
+            ? `The last pressed key is [${this.state.key}]`
+            : 'Nothing was pressed yet'}
+        </p>
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.setState, false);
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.key
```
makandriival, line 26: you could restructure this.state.key to make it look nicer

src/App.tsx
```diff
@@ -1,7 +1,33 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string;
```
makandriival, line 4: i did it the same way as you (only i used 'nothing' for default key value). Its probably mistake and we should use string | null in types and null for Key default value

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (e) => {
```
makandriival, line 13: couple tusks ago i got a comment about naming variable inside function with one later 'e', 'n' etc. not the best practice

makandriival (approved):
GJ! I left some comments, hope you gonna read them

src/App.tsx
SerhiiKirik: Agree, do it

src/App.tsx
SerhiiKirik: In a lot of key `e` - it's ok, but let's put `event` name here

src/App.tsx
SerhiiKirik: Yeah, just make a destructuring from this.state

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+      this.setState({ key: e.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.setState, false);
```
SerhiiKirik, line 19: Do you need this `false` here?

SerhiiKirik (changes requested):
Well done 🚀 
Also, do not ignore student's comments, answer or make a changes 🌍

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string | null;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    key: '',
```
SerhiiKirik, line 9:
```suggestion
    key: null,
```

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+    document.removeEventListener('keyup', this.setState);
+  }
+
+  render() {
+    const { key } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.key
```
SerhiiKirik, line 28:
```suggestion
          {key
```

SerhiiKirik (approved):
Well done 🚀

PR #699, opened 2023-04-12
keyboard clicker added

DenysKrysanov:
[DEMO LINK](https://DenysKrysanov.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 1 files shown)


Olekk17 (approved):

src/App.tsx
```diff
@@ -1,7 +1,38 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
```

SerhiiKirik (approved):
Great 🚀

PR #700, opened 2023-04-12
complete keyboard

VenomMB:
[DEMO LINK](https://VenomMB.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKey);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKey);
+  }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+  handleKey = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">{ pressedKey !== '' ? `The last pressed key is [${pressedKey}]` : 'Nothing was pressed yet'}</p>
```
KosolapovAnatolii, line 29: also you can do without the expression " pressedKey !== '' "
```suggestion
        <p className="App__message">{ pressedKey 
          ? `The last pressed key is [${pressedKey}]`
           : 'Nothing was pressed yet'}
         </p>
```

KosolapovAnatolii (approved):
Absolutly fantastic solution !

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    pressedKey: '',
```

src/App.tsx
SerhiiKirik: Agree, do it

SerhiiKirik (changes requested):
Well done 🚀 

SerhiiKirik (approved):
Great 🔥

PR #704, opened 2023-04-13
Solution

KosolapovAnatolii:
[DEMO LINK](https://KosolapovAnatolii.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,19 +1,19 @@
 import { Component } from 'react';
 
 type State = {
-  pressedKey: string;
+  pressedKey: string | null;
 };
 
 export class App extends Component<{}, State> {
   state: Readonly<State> = {
-    pressedKey: '',
+    pressedKey: null,
   };
 
   componentDidMount() {
     document.addEventListener('keyup', this.handleDalateClick);
   }
 
-  componentWillUnmount(): void {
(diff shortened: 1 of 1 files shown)


yuliia-anisimova (approved):
LGTM!

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: '',
```

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleDalateClick);
+  }
+
+  componentWillUnmount(): void {
```
SerhiiKirik, line 16: It's extra, bc TypeScript already knows about React types
Also,  choose just 1 way & use it, bc in 12 lines you don't do this
```suggestion
  componentWillUnmount() {
```

SerhiiKirik (changes requested):
Well done 🚀 
Let's improve your code 🧑‍💻

PR #706, opened 2023-04-13
finish

Rage1502:
[DEMO LINK](https://Rage1502.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey : string | null;
+};
+
+export class App extends Component<{}, State> {
+  state: State = {
+    pressedKey: null,
+  };
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    document.addEventListener('keyup', this.keyPresser);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.keyPresser);
+  }
+
+  keyPresser = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
```
Kavon-Skela, line 22: I think it is better to write a type of function result. Even if function return nothing

Kavon-Skela, line 22: Checklist p.1

Kavon-Skela (changes requested):
Good job! Lets make some improvements according to the checklist

src/App.tsx
Rage1502: We don't have a checklist for this task.

Kavon-Skela: image

SerhiiKirik (approved):
LGTM 🔥

PR #707, opened 2023-04-13
created class component App

Sbavdik:
[DEMO LINK](https://Sbavdik.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | null;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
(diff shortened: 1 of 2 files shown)


Nazarii98 (approved):
Good job!

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | null;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
```
SerhiiKirik, line 9: So, you put a type `string | null`
Where do you set null?)
```suggestion
    pressedKey: null,
```
Let's do it here

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  };
+
+  componentDidMount(): void {
```
SerhiiKirik, line 12: It's extra, bc TypeScript already knows about React types
```suggestion
  componentDidMount() {
```

SerhiiKirik (approved):
Well done 🚀

PR #708, opened 2023-04-13
Solution

yuliia-anisimova:
[DEMO LINK](https://yuliia-anisimova.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
+  };
+
+  componentDidMount() {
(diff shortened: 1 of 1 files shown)


MylaCrystal (approved):
Good job!

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
```

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyUp);
+  }
+
+  handleKeyUp = (event: KeyboardEvent) => (this.setState({ key: event.key }));
```
SerhiiKirik, line 20:
```suggestion
  handleKeyUp = (event: KeyboardEvent) => (
    this.setState({ key: event.key })
  );
```

PR #709, opened 2023-04-13
Solution

Anna-Kuzmenko:


Anna-Kuzmenko:
[DEMO LINK](https://Anna-Kuzmenko.github.io/react_keyboard/)

src/App.tsx
```diff
@@ -1,7 +1,39 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
```
vksazonov, line 5: I'm not sure, but maybe it's better to write down string | null nad in pressedKey to write down null

vksazonov (approved):

src/App.tsx
SerhiiKirik: Agree

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
```
SerhiiKirik, line 14: DRY
Move this callback to separate func

SerhiiKirik (changes requested):
Well done 🚀 
Let's improve your code 👩‍💻

src/App.tsx
Anna-Kuzmenko: Moved )

PR #714, opened 2023-04-14
Solution

vksazonov:
[DEMO LINK](https://vksazonov.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,7 +16,7 @@ export class App extends React.Component<{}, State> {
   }
 
   componentWillUnmount(): void {
-    document.addEventListener('keyup', (event: KeyboardEvent) => {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
       this.setState({ key: event.key });
     });
   }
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ key: event.key });
+    });
+  }
+
+  componentWillUnmount(): void {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
```
kukla1989, line 19: here you should remove event listener with "removeEventListener" instead  of add another event listener
2. [CODE KNOWLEDGE] - Check if you remove your event listener, when component will unmount from page.

kukla1989 (changes requested):
good job! but be more carefully with checklist

yapelenua (approved):
Good job)

liliia-livandovskaya (approved):

PR #715, opened 2023-04-14
make class App

Vlad-Lischenovskiy:
[DEMO LINK](https://Vlad-Lischenovskiy.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  pressedKey: string | null;
+}
+
+export class App extends React.Component {
+  state: Readonly<State> = {
+    pressedKey: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyPress);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleKeyPress);
+  }
+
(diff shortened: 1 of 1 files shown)


TanyaS133 (approved):
Great!)

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  pressedKey: string | null;
+}
+
+export class App extends React.Component {
```
liliia-livandovskaya, line 7: Since you are using TypeScript, it's a good practice to define the Props type for the component. You can define an empty interface if your component doesn't have any props. For example, `interface Props {}` and then use it like `class App extends React.Component<Props, State>`.

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.pressedKey
+            ? `The last pressed key is [${this.state.pressedKey}]`
+            : 'Nothing was pressed yet'}
+        </p>
+      </div>
```
liliia-livandovskaya, line 34:
```suggestion
  render() {
    const { pressedKey } = this.state;
    
    return (
      <div className="App">
        <p className="App__message">
          {pressedKey
            ? `The last pressed key is [${pressedKey}]`
            : 'Nothing was pressed yet'}
        </p>
      </div>
```

liliia-livandovskaya (approved):

PR #716, opened 2023-04-14
Develop

Anastasia2403:
[DEMO LINK](https://Anastasia2403.github.io/react_keyboard/)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+    pressedKey: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyUp);
+  }
```
Goto-Jesus, line 18: It is not a mistake, but try to keep one style of code... if you give a type for methods do this always or never)
I mean look at 12 and 16 strings.. : void {}

Goto-Jesus (approved):
Good job

src/App.tsx
SerhiiKirik: agree, just remove :void, bc typescript already know about this type

SerhiiKirik (approved):
Well done 🚀

PR #717, opened 2023-04-14
Class component

roshirin:
[DEMO LINK](https://roshirin.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,31 +1,37 @@
 import React from 'react';
 
 interface State {
-  pressedKey: string;
+  pressedKey: string | null;
 }
 
 export class App extends React.Component<{}, State> {
   state = {
-    pressedKey: 'Nothing was pressed yet',
+    pressedKey: null,
   };
 
-  componentDidMount(): void {
+  componentDidMount() {
     window.addEventListener('keydown', this.handleKeyboardPress);
   }
 
-  componentWillUnmount(): void {
+  componentWillUnmount() {
     window.removeEventListener('keydown', this.handleKeyboardPress);
   }
 
-  handleKeyboardPress = (event: KeyboardEvent): void => {
-    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
+  handleKeyboardPress = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
   };
 
   render() {
+    const { pressedKey } = this.state;
+
     return (
       <div className="App">
         <p className="App__message">
-          {this.state.pressedKey}
+          {
+            pressedKey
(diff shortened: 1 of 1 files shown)


StasLukianchuk (approved):
GJ!

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  pressedKey: string;
+}
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
```suggestion
    pressedKey: null,
```

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+  };
+
+  componentDidMount(): void {
```
SerhiiKirik, line 12: It's extra, bc TypeScript already knows about React types
```suggestion
  componentDidMount() {
```

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+    window.addEventListener('keydown', this.handleKeyboardPress);
+  }
+
+  componentWillUnmount(): void {
+    window.removeEventListener('keydown', this.handleKeyboardPress);
+  }
+
+  handleKeyboardPress = (event: KeyboardEvent): void => {
+    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
```
SerhiiKirik, line 21: So, each time you set this string?
So, why do you give the name `pressedKey` for this string, it can be `Nothing was pressed yet`?
Just put a key into the state

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+
+  componentDidMount(): void {
+    window.addEventListener('keydown', this.handleKeyboardPress);
+  }
+
+  componentWillUnmount(): void {
+    window.removeEventListener('keydown', this.handleKeyboardPress);
+  }
+
+  handleKeyboardPress = (event: KeyboardEvent): void => {
```
SerhiiKirik, line 20:
```suggestion
  handleKeyboardPress = (event: KeyboardEvent) => {
```

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.pressedKey}
```
SerhiiKirik, line 28: you can make a destructuring here, from this.state

SerhiiKirik (changes requested):
Well done 🚀 

SerhiiKirik (approved):
LGTM 🔥

PR #718, opened 2023-04-14
react keybord

glots-a:
[demo](https://glots-a.github.io/react_keyboard)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,53 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  wasPresed: boolean;
+  keyPres: null | string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    wasPresed: false,
+    keyPres: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', (e: KeyboardEvent) => {
+      this.setState({
+        wasPresed: true,
+        keyPres: e.key,
+      });
+    });
+  }
+
+  componentWillUnmount(): void {
+    document.addEventListener('keyup', (e: KeyboardEvent) => {
+      this.setState({
+        wasPresed: true,
+        keyPres: e.key,
+      });
+    });
+  }
+
+  // handlePressKey(e: KeyboardEvent) {
+  //   this.setState({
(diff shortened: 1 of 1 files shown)


DimaDvm (approved):

DimaDvm:
Gj!

src/App.tsx
```diff
@@ -1,7 +1,42 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  wasPresed: boolean;
```
SerhiiKirik, line 4: Why do you need this?
If you have a null in `keyPres` you know that no one pressed a button, no?
```suggestion
```

src/App.tsx
```diff
@@ -1,7 +1,42 @@
...
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    wasPresed: false,
+    keyPres: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handlePressKey.bind(this));
```
SerhiiKirik, line 15:
```suggestion
    document.addEventListener('keyup', this.handlePressKey);
```

src/App.tsx
```diff
@@ -1,7 +1,42 @@
...
+  keyPres: null | string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    wasPresed: false,
+    keyPres: null,
+  };
+
+  componentDidMount(): void {
```
SerhiiKirik, line 14: It's extra, bc TypeScript already knows about React types

```suggestion
  componentDidMount() {
```

src/App.tsx
```diff
@@ -1,7 +1,42 @@
...
+      keyPres: e.key,
+    });
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
+            this.state.wasPresed
```
SerhiiKirik, line 34: you can make a destructuring from this.state in render

SerhiiKirik (changes requested):
Great 🔥 
Let's make some improvements 🧑‍💻

src/App.tsx
SerhiiKirik: It's just callback. So, your code doesn't work without bind?

SerhiiKirik (approved):
Well done 🚀

PR #719, opened 2023-04-14
Develop

kukla1989:
[DEMO LINK](https://kukla1989.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,41 @@
 import React from 'react';
 
+interface State {
+  key: string;
+}
+
+export class WhatIsPressed extends React.Component<{}, State> {
+  state = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  pressedKey: string;
+}
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: 'Meta',
```
SerhiiKirik, line 9: why meta?
Just store null here if you don't have a value yet

SerhiiKirik (approved):
Nice 🚀

PR #722, opened 2023-04-14
solution

artem-kyrylenko:
[DEMO LINK](https://artem-kyrylenko.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  pressedKey: string;
+}
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    window.addEventListener('keyup', (e: KeyboardEvent) => {
+      this.setState({ pressedKey: e.key });
+    });
+  }
+
+  render() {
+    const { pressedKey } = this.state;
+
+    const infoMessage = pressedKey
+      ? `The last pressed key is [${this.state.pressedKey}]`
+      : 'Nothing was pressed yet';
+
+    return (
+      <div className="App">
+        <p className="App__message">{infoMessage}</p>
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  componentDidMount() {
+    window.addEventListener('keyup', (e: KeyboardEvent) => {
+      this.setState({ pressedKey: e.key });
+    });
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('keyup', (e: KeyboardEvent) => {
+      this.setState({ pressedKey: e.key });
```
VladislavaT, line 20: mb it`s better to write 'event' instead of 'e'

VladislavaT (approved):
GJ!

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  pressedKey: string;
+}
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
```
SerhiiKirik, line 9: It would be better if you put null here. It's mean that you don't have a value now

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  };
+
+  componentDidMount() {
+    window.addEventListener('keyup', (e: KeyboardEvent) => {
+      this.setState({ pressedKey: e.key });
```
SerhiiKirik, line 14: move this callback to the separate func

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  componentWillUnmount() {
+    window.removeEventListener('keyup', (e: KeyboardEvent) => {
+      this.setState({ pressedKey: e.key });
+    });
+  }
+
+  render() {
+    const { pressedKey } = this.state;
+
+    const infoMessage = pressedKey
```
SerhiiKirik, line 27: It's unnecessary to create separate variable for this

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+      ? `The last pressed key is [${this.state.pressedKey}]`
```
SerhiiKirik, line 28: You already destructured this value from this.state
```suggestion
      ? `The last pressed key is [${pressedKey}]`
```

SerhiiKirik (changes requested):
Great 🔥 
Let's make some improvements 🧑‍💻

SerhiiKirik (approved):
LGTM 🔥

PR #726, opened 2023-04-14
Solution for test

DimaDvm:
DEMO LINK - https://DimaDvm.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,38 @@
+// import { prependOnceListener } from 'process';
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    });
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
+            this.state.pressedKey
+              ? `The last pressed key is [${this.state.pressedKey}]`
```
liliia-livandovskaya, line 30:
```suggestion
  render() {
  const { pressedKey } = this.state;
  
    return (
      <div className="App">
        <p className="App__message">
          {
            this.state.pressedKey
              ? `The last pressed key is [${pressedKey}]`
```

liliia-livandovskaya (approved):

PR #734, opened 2023-04-17
solution

Spec0s:
https://spec0s.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string | null;
+};
+
+export class App extends React.Component<State> {
+  state = {
+    pressedKey: null,
+  };
+
+  componentDidMount() {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
```
qoa11a, line 22: Functions are objects in JavaScript, and objects are stored by reference. It means that your callback function in the 13th row isn't the same as the one in the 19th row, as a result, this event listener won't be removed.
To fix it, you should simply create a function and pass it as a callback in both methods `addEventListener` and `removeEventListener`.

https://javascript.info/object-copy
https://javascript.info/introduction-browser-events#addeventlistener

image

qoa11a (changes requested):
Nice and clean code! 👍 

Though, you have one problem which I've commented upon. Also, the demo link doesn't work. Please, run `npm run deploy` again and check if it's working.

anstsot (approved):
Great job, approved!

PR #742, opened 2023-04-22
first commit

haduigon:
https://haduigon.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,38 @@
 import React from 'react';
+import { KeyPressed } from './Component';
+import './App.scss';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  click: boolean,
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    click: true,
+  };
+
+  render() {
+    const { click } = this.state;
+    const clickDestroy = () => {
+      this.setState({ click: false });
+    };
+
+    const clickCreate = () => {
+      this.setState({ click: true });
+    };
+
+    return (
+      <>
+        <div className="App">
+          <p className="App__message">
+            { click && <KeyPressed /> }
+          </p>
+        </div>
+        <button type="submit" onClick={clickDestroy}>Destroy element</button>
+        <button type="submit" onClick={clickCreate}>Create element</button>
+      </>
+    );
+  }
+}
+
+export default App;
--- a/src/Component.tsx
+++ b/src/Component.tsx
@@ -1,0 +1,42 @@
+import { Component } from 'react';
+
+export class KeyPressed extends Component {
+  state = {
+    keyPressed: '',
+  };
+
+  componentDidMount() {
+    setTimeout(() => {
+      this.setState({ keyPressed: 'Nothing was pressed' });
+    }, 1000);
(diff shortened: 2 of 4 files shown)


src/App.tsx
```diff
@@ -1,7 +1,38 @@
...
+
+    return (
+      <>
+        <div className="App">
+          <p className="App__message">
+            { click && <KeyPressed /> }
+          </p>
+        </div>
+        <button type="submit" onClick={clickDestroy}>Destroy element</button>
+        <button type="submit" onClick={clickCreate}>Create element</button>
```
TarasHoliuk, line 32: `submit` is a wrong type here
anyway, why do you need these buttons here?

src/Component.tsx
```diff
@@ -0,0 +1,42 @@
+import { Component } from 'react';
+
+export class KeyPressed extends Component {
+  state = {
+    keyPressed: '',
+  };
+
+  componentDidMount() {
+    setTimeout(() => {
+      this.setState({ keyPressed: 'Nothing was pressed' });
```
TarasHoliuk, line 10: you shouldn't do it in the `componentDidMount` with a timeout
actually, you shouldn't do it at all ;)

src/Component.tsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  inform = () => {
+    document.removeEventListener('keyup', this.keyBoardHandler);
+  };
+
+  keyBoardHandler = (event?: KeyboardEvent) => {
+    const message = `The last pressed key is ${event?.key}`;
+
+    this.setState({ keyPressed: message });
```
TarasHoliuk, line 30: don't save the whole message, variable name is `keyPressed`, so save only `event.key` there

src/Component.tsx
TarasHoliuk: should be `Nothing was pressed yet` (to pass tests)

src/Component.tsx
```diff
@@ -0,0 +1,42 @@
...
+  };
+
+  render() {
+    const message2 = this.state.keyPressed;
+
+    return (
+      <>
+        {message2}
```
TarasHoliuk, line 38: use conditional rendering here
keyPressed is an empty string, render the corresponding message, otherwise render string using interpolation with `keyPressed` variable (the similar to the `message` variable on the 28 row)

TarasHoliuk (changes requested):

haduigon:
have rewrote all )

src/Component.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup',
+      (event: KeyboardEvent) => this.keyBoardHandler(event));
+  }
+
+  componentDidUpdate() {
+    document.addEventListener('keyup',
+      (event: KeyboardEvent) => this.keyBoardHandler(event));
+  }
```
anstsot, line 16: you don't need to set event listener in `componentDidUpdate`. it will work if you set it only in `componentDidMount`
```suggestion
  componentDidMount() {
    document.addEventListener('keyup',
      (event: KeyboardEvent) => this.keyBoardHandler(event));
  }
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

TarasHoliuk (approved):

PR #749, opened 2023-04-26
solution

Dka888:
[DEMO LINK](https://Dka888.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: boolean,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    key: false,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleClick);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleClick);
+  }
+
+  handleClick = () => {
+    this.setState({ key: true });
+  };
+
+  render() {
+    const condition = this.state.key;
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: boolean,
+};
```
ghostargg, line 5: your key variable have to be string or null, when you initialize it set it to null then change it to result of handleClick function (event.key), right now your solution doesnt work as intended

ghostargg (changes requested):

src/App.tsx
```diff
@@ -28,7 +28,7 @@ export class App extends React.Component<{}, State> {
       <div className="App">
         <p className="App__message">
           {condition
-            ? 'The last pressed key is [Enter]'
+            ? `The last pressed key is [${condition}]`
```
ghostargg, line 31: change condition to ${this.state.key} or even better destruct state in your render and use just ${key}

ghostargg (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleClick);
+  }
+
+  handleClick = (event: KeyboardEvent) => {
+    this.setState({ key: event.key });
+  };
+
+  render() {
+    const condition = this.state.key;
```
d-wasilewski, line 25: Why do you call this condition?
```suggestion
    const { key } = this.state;
```

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #755, opened 2023-04-30
Solution-1

MykolaButylkov:
[DEMO LINK](https://MykolaButylkov.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
+import { Key } from './component/Component';
 
 export const App: React.FC = () => (
   <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
+    <Key key={0} />
   </div>
 );
--- a/src/component/Component.tsx
+++ b/src/component/Component.tsx
@@ -1,0 +1,21 @@
+import { Component, ReactNode } from 'react';
+
+export class Key extends Component {
+  state = {
+    message: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
+import { Key } from './component/Component';
 
 export const App: React.FC = () => (
   <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
+    <Key key={0} />
```
etojeDenys, line 6: The 'key' prop should not be used here because it's a reserved prop in React and its purpose is to help with list rendering. Use a different prop name to pass the last pressed key value.

src/component/Component.tsx
```diff
@@ -0,0 +1,21 @@
+import { Component, ReactNode } from 'react';
+
+export class Key extends Component {
```
etojeDenys, line 3: Consider extending React.Component with appropriate TypeScript generic, such as React.Component<{}, StateType> to ensure type safety.

src/component/Component.tsx
```diff
@@ -0,0 +1,21 @@
...
+  state = {
+    message: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
```
etojeDenys, line 8: Use componentDidMount to add a keyup handler, but it's missing componentWillUnmount to remove the handler.

etojeDenys (changes requested):

OleziO (approved):

PR #756, opened 2023-05-02
Solution

Vaskovics:
<a href='https://vaskovics.github.io/react_keyboard/'>Demo</a>

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
-import React from 'react';
+import React, { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+  componentDidMount(): void {
+    document.documentElement.addEventListener('keydown', this.keyHandler);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keydown', this.keyHandler);
+  }
+
+  keyHandler = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
```
anstsot, line 21: it would be better to store the `event.key` only and add check in render method
```suggestion
    this.setState({ pressedKey: event.key });
```

anstsot (approved):
Good job, approved!

PR #757, opened 2023-05-02
solution

danielpopek94:
[DEMO LINK](https://danielpopek94.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string | undefined
+}
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    key: undefined,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleClick);
+  }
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string | undefined
+}
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
```
d-wasilewski, line 8: TS will infer this type from initial object
```suggestion
  state = {
```

d-wasilewski: also rename the variable to `keyPressed` as per task description

src/App.tsx
```diff
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string | undefined
```
d-wasilewski, line 4: Undefined means a variable has been declared but has not been assigned a value, while null is an assignment value set by a programmer. If you set it manually use null
```suggestion
  key: string | null
```

d-wasilewski (changes requested):

PR #759, opened 2023-05-04
solution

MikiOOOO:
https://MikiOOOO.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,7 +4,6 @@ interface State {
   pressedKey: string | null;
 }
 
-// eslint-disable-next-line
 export class App extends React.Component<{}, State> {
   state = {
     pressedKey: null,
@@ -31,7 +30,7 @@ export class App extends React.Component<{}, State> {
           {
             pressedKey
               ? `The last pressed key is [${pressedKey}]`
-              : 'Nothing was pressed yet.'
+              : 'Nothing was pressed yet'
           }
         </p>
       </div>


src/App.tsx
```diff
@@ -1,7 +1,38 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
```
sylcym, line 3: I think, you can use `interface `here instead of `type`

src/App.tsx
```diff
@@ -1,7 +1,38 @@
...
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
```
sylcym, line 24: I think, you can move  `(event: KeyboardEvent) => { this.setState({ pressedKey: event.key })`  to a function and use it in `componentDidMount `and in `componentWillUnmount `like this:

```
componentDidMount() {
  document.addEventListener('keyup', this.handleKeyUp);
}

componentWillUnmount() {
  document.removeEventListener('keyup', this.handleKeyUp);
}

handleKeyUp = (event: KeyboardEvent) => {
};
```

sylcym (changes requested):

src/App.tsx
d-wasilewski: Yes, use interfaces to describe objects

src/App.tsx
```diff
@@ -1,7 +1,38 @@
...
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
+            this.state.pressedKey
+              ? `The last pressed key is [${this.state.pressedKey}]`
```
d-wasilewski, line 31: you use it twice, you can destructure it

d-wasilewski (changes requested):
try to fix the tests

src/App.tsx
MikiOOOO: I cannot fix tests, I tried reinstalling npm with cleaning the cache beforehand and it is still not working.

d-wasilewski: It's not a local problem, they don't run on your machine here

d-wasilewski: There is a problem with your code. Subtle, but check what the failed tests say
image

PR #766, opened 2023-05-07
solution

YBM3:
[DEMO LINK](https://YBM3.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,39 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  message: string | null;
+}
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    message: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keypress', this.handleButtonPress);
+  }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  message: string | null;
```
bjeziorek, line 4: I'm not sure if message is right name for this, maybe something like pressedKey would be better :).

bjeziorek (approved):
In general code is fine. Just consider changing state name.

d-wasilewski (changes requested):
Fix the tests

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface AppState {
+  pressedKey: string | null;
+}
+
+export class App extends Component<{}, AppState> {
+  state: Readonly<AppState> = {
```
d-wasilewski, line 8: TS will infer it from initial object
```suggestion
  state = {
```

d-wasilewski (approved):

PR #767, opened 2023-05-07
commit message

EvgenyNikitinEvgeny:
https://EvgenyNikitinEvgeny.github.io/react_keyboard/


src/App.tsx
```diff
@@ -1,7 +1,30 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type Message = {
```
SanyaBratashchuk, line 3:
```suggestion
type State = {
```

src/App.tsx
```diff
@@ -1,7 +1,30 @@
...
+  message: string,
+};
+
+export class App extends Component<{}, Message> {
+  state = {
+    message: 'Nothing was pressed yet',
+  };
```
SanyaBratashchuk, line 10:
```suggestion
  state = {
    message: null,
  };
```
It is more flexible

SanyaBratashchuk (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,32 @@
...
+    });
+  }
+
+  render() {
+    const { key } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {key === null ? 'nothing was pressed yet' : key }
```
DarkMistyRoom, line 27:
```suggestion
          {key === null ? 'Nothing was pressed yet' : key }
```

DarkMistyRoom: Please check the test results before sending the PR 
image

DarkMistyRoom (approved):
Good but try to pass all tests

PR #770, opened 2023-05-07
Solution

KielpinskiJ:
[DEMO LINK](https://KielpinskiJ.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleKeyUp);
+  }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
```
d-wasilewski, line 3: use TS to type the state
`export class App extends React.Component<{}, State> {`

d-wasilewski (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,13 @@
...
-export class App extends React.Component {
-  state = {
+type Props = {};
+
+interface State {
+  pressedKey: string | null;
+}
+
+export class App extends React.Component<Props, State> {
+  state: Readonly<State> = {
```
d-wasilewski, line 10: TS will infer it here from initial object, you can safely go with `state = {`

d-wasilewski (changes requested):

src/App.tsx
KielpinskiJ: I think Volodymyr told us to use `state: Readonly<State> = {}` because it provides a higher level of safety than just `state = {}`.

d-wasilewski: Yes, my bad here

d-wasilewski (approved):

PR #771, opened 2023-05-07
Solution

SuaBy514:
[DEMO LINK](https://suaby514.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string | null,
+}
+
+export class App extends React.Component<{}, State> {
+  state = {
+    key: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyPress);
+  }
+
+  handleKeyPress = (event: KeyboardEvent) => {
+    this.setState({ key: event.key });
+  };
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string | null,
+}
+
+export class App extends React.Component<{}, State> {
+  state = {
+    key: null,
```
d-wasilewski, line 9: > Make the App a class component with pressedKey in the state.
```suggestion
    pressedKey: null,
```

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    document.addEventListener('keyup', this.handleKeyPress);
+  }
+
+  handleKeyPress = (event: KeyboardEvent) => {
+    this.setState({ key: event.key });
+  };
+
+  componentDidUnmount(): void {
+    document.removeEventListener('keyup', this.handleKeyPress);
+  }
```
d-wasilewski, line 22: Try to organize similar things
```suggestion
componentDidUnmount(): void {
    document.removeEventListener('keyup', this.handleKeyPress);
  }

  handleKeyPress = (event: KeyboardEvent) => {
    this.setState({ key: event.key });
  };
```

d-wasilewski (changes requested):

src/App.tsx
SuaBy514: Works with componentWillUnmount though

PR #777, opened 2023-05-09
add solution

PufAnimalsDev:
[DEMO LINK](https://pufanimalsdev.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string;
+}
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string;
+}
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
```
d-wasilewski, line 8: TS will infer it from initial object, you can safely go with this
```suggestion
  state = {
```

d-wasilewski: also change `key` to `keyPressed` as per task description

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+    key: 'Nothing was pressed yet',
```
d-wasilewski, line 9: well done with setting this as an initial value 🚀

d-wasilewski (approved):

PR #779, opened 2023-05-09
Solution

NadieinOleh:
https://NadieinOleh.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
+import { PressedKey } from './Components/PressedKey';
 
 export const App: React.FC = () => (
   <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
+    <PressedKey />
   </div>
 );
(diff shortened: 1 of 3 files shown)


src/Components/PressedKey.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+type State = {
+  key: string;
+};
+
+export class PressedKey extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    key: '',
```
polosanya, line 9:
```suggestion
    pressedKey: '',
```
image

src/App.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
+import { PressedKey } from './Components/PressedKey';
 
 export const App: React.FC = () => (
   <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
+    <PressedKey />
```
polosanya, line 6: There is no need to create a new component. You can store all logic in App.tsx according to the task description

polosanya (changes requested):
Almost done!

OleziO (approved):

PR #794, opened 2023-05-20
all tests good

Mikhaylo-Orlov:
[DEMO LINK](https://Mikhaylo-Orlov.github.io/react_keyboard/)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends Component {
```
TarasHoliuk, line 3: do not overwrite it using the class component - functional components are much more relevant. Solve your tasks with FC.

TarasHoliuk (changes requested):

PR #799, opened 2023-05-26
done

generatedman:
[DEMO LINK](https://generatedman.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ key: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ key: event.key });
+    });
+  }
+
+  render() {
+    return (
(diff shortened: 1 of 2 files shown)


DidenkoDenis306 (changes requested):
It would be cool to use typescipt in this task)))

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
```
DidenkoDenis306, line 9: Handle callbacks to separate function inside you component.

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+      this.setState({ key: event.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
```
DidenkoDenis306, line 15: Same here

DidenkoDenis306 (changes requested):

DidenkoDenis306 (changes requested):
Good job, a few moments to fix

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+      this.setState({ key: event.key });
+    });
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.key === ''
```
33Orange, line 24: What would be if `key === undefine | null` ? Change condition

```suggestion
          {this.state.key
```

33Orange (approved):
Well done

PR #802, opened 2023-05-29
add task solution

maksym-nemera:
[DEMO LINK](https://maksym-nemera.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,15 +5,17 @@ export class App extends React.Component {
     pressedKey: 'Nothing was pressed yet',
   };
 
-  componentDidMount(): void {
-    document.addEventListener('keyup', (event: KeyboardEvent) => {
-      this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
-    });
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKeyUp);
   }
 
-  componentWillUnmount(): void {
-    document.removeEventListener('keyup', this.componentDidMount);
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyUp);
   }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,25 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
```
dinaminahi, line 9: let's create a separate function for event

src/App.tsx
```diff
@@ -1,7 +1,25 @@
...
+      this.setState({ pressedKey: `The last pressed key is [${event.key}]` });
+    });
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.componentDidMount);
```
dinaminahi, line 15: and use it here instead of lifecycle method

dinaminahi (changes requested):
Good job 👍 
but let's fix few moments

src/App.tsx
```diff
@@ -1,7 +1,25 @@
...
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount(): void {
```
dinaminahi, line 8: No need of specifying return `void` type, typescript can handle this by itself

silchyn (approved):

PR #803, opened 2023-05-29
Develop

RomanHrytsiuk:
https://RomanHrytsiuk.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: 'Nothing was pressed yet',
```
TarasHoliuk, line 9: you save not only `pressedKey` but the entire message
since the state value is named `pressedKey` (as required in the task), you should save there only `event.key` (empty string by default)

then inside the render method, inside `App__message` use ternary operator to show the correct message depending on `pressedKey` value: https://react.dev/learn/conditional-rendering#conditional-ternary-operator--

TarasHoliuk (changes requested):

SanyaBratashchuk (approved):

PR #805, opened 2023-06-01
ClassComponent

IvanFront-end:
[DEMO LINK](https://IvanFront-end.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  messageNothing: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+    document.addEventListener('keyup', this.handler);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handler);
+  }
+
+  handler = (event: KeyboardEvent) => {
+    this.setState({
+      messageNothing: `The last pressed key is [${event.key}]`,
```
TarasHoliuk, line 22: you set `message` to property with key `messageNothing` - a bit inconsistently)

do it in this way: create `pressedKey` property in the state (as required) and save there **only** `event.key` (empty string by default)
then, using [conditional rendering/ternary operator](https://react.dev/learn/conditional-rendering) render the message you need

TarasHoliuk (changes requested):

TarasHoliuk (approved):
:+1:

PR #809, opened 2023-06-03
done

VladyslavKucher:
https://VladyslavKucher.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,39 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  selectedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    selectedKey: 'Nothing was pressed yet',
+  };
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  selectedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    selectedKey: 'Nothing was pressed yet',
```
witflash, line 9: It's better set `null` if there is no pressed key.
And message should be rendered based on this nullable value

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+  };
+
+  render() {
+    const currentKey = this.state.selectedKey;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
+            (currentKey === 'Nothing was pressed yet')
```
witflash, line 31:
```
currentKey === null
  ? 'Nothing was pressed yet'
  : 'The last....'
```

witflash (changes requested):
Good job!
Could you make a small improvement?

OleziO (approved):

PR #816, opened 2023-06-08
add task solution

EugeneSherepa:
[DEMO LINK](https://EugeneSherepa.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,15 @@
-import React from 'react';
+import { Component } from 'react';
 
-export class App extends React.Component {
+export class App extends Component {
   state = {
     pressedKey: null,
   };
 
-  componentDidMount(): void {
+  componentDidMount() {
     document.addEventListener('keyup', this.handleKeyUp);
   }
 
-  componentWillUnmount(): void {
+  componentWillUnmount() {
     document.removeEventListener('keyup', this.handleKeyUp);
   }
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,11 @@
 import { Component } from 'react';
 
-export class App extends Component {
-  state = {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends Component {
+  state = {
```
ruslanpashkov, line 4: add types to the state to make sure it will never be changed from the outside

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyUp);
+  }
+
+  handleKeyUp = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
```
ruslanpashkov, line 17: what if the pressed key will be space? let's fix this case

ruslanpashkov (changes requested):

src/App.tsx
ruslanpashkov: also type your class component*

src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+
+    if (this.state.pressedKey === ' ') {
```
silchyn, line 23: you can't be sure that state is actual at this point

silchyn (changes requested):

33Orange (approved):
Well done

PR #822, opened 2023-06-08
add solution

growingananas:
[DEMO LINK](https://growingananas.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,47 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  event: string | null;
+};
+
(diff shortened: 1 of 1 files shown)


ruslanpashkov (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,40 @@
...
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
+            pressedKey
+              ? `The last pressed key is [${pressedKey === ' ' ? 'Space' : pressedKey}]`
```
silchyn, line 32: this should be handled at the moment of storing it

silchyn (changes requested):

src/App.tsx
growingananas: Sending another update. Hope, this time it will be all right)

silchyn (approved):

PR #827, opened 2023-06-08
solution

okosohor:


src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {
+            !this.state.pressedKey
+              ? 'Nothing was pressed yet'
+              : `The last pressed key is [${this.state.pressedKey}]`
+          }
```
naziamoff, line 33: - no need to add spacing here. The code doesn't get any clearer, but you lose in ergonomics a lot
- let's start conditional rendering with positive condition - it is easier to read
```suggestion
          {!this.state.pressedKey
              ? 'Nothing was pressed yet'
              : `The last pressed key is [${this.state.pressedKey}]`}
```

src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
+        </p>
+      </div>
+    );
+  }
+}
+
+// render() {
```
naziamoff, line 40: never leave comments in your code unless you really need to

naziamoff (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State =
+{
+  pressedKey: string | null,
+};
+
```
naziamoff, line 7:
```suggestion
type State = {
  pressedKey: string | null,
};

```

src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+  handleDocumentClick = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          { this.state.pressedKey
```
naziamoff, line 29:
```suggestion
          {this.state.pressedKey
```

naziamoff (approved):

PR #829, opened 2023-06-08
add solution

serentinos:
[DEMO LINK](https://serentinos.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string;
+}
+
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
+  };
+
+  componentDidMount(): void {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  key: string;
+}
+
+export class App extends Component<{}, State> {
+  state = {
+    key: '',
```
33Orange, line 9: Use `null` instead empty string for initial empty value

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.keyboardHandler);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.keyboardHandler);
+  }
+
+  keyboardHandler = (event: KeyboardEvent) => {
```
33Orange, line 20:
```suggestion
  handleKeyUp= (event: KeyboardEvent) => {
```

33Orange (changes requested):
Almost done

A few moments to fix

naziamoff (approved):
awesome 🔥

PR #830, opened 2023-06-08
initial commit

44shaggy:
[DEMO LINK](https://44shaggy.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,41 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type Props = {};
+
+type State = {
+  pressedKey: string | null,
+};
+
+export class App extends React.Component<Props, State> {
+  state = {
+    pressedKey: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+};
+
+export class App extends React.Component<Props, State> {
+  state = {
+    pressedKey: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
```
Anastasiia-Svintsova, line 16: It's better to create a separate function for this action to not repeat the same code twice

Anastasiia-Svintsova (changes requested):
Good job! Let's improve it a little bit

src/App.tsx
44shaggy: makes sense

src/App.tsx
```diff
@@ -1,7 +1,41 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type Props = {};
```
lerastarynets, line 3: Props is an empty type, it can be removed and the component can be defined as `React.Component<{}, State>`.

lerastarynets (approved):
Excellent work! Just one tiny thing to be improved)

src/App.tsx
44shaggy: Oh, it's aproved already, no need to re-review :)

PR #833, opened 2023-06-08
solution

dimoshen:
[DEMO LINK](https://dimoshen.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,14 +10,14 @@ export class App extends Component<{}, State> {
   };
 
   componentDidMount() {
-    document.addEventListener('keyup', this.keyStateHandler);
+    document.addEventListener('keyup', this.handleKeyState);
   }
 
   componentWillUnmount() {
-    document.removeEventListener('keyup', this.keyStateHandler);
+    document.removeEventListener('keyup', this.handleKeyState);
   }
 
-  keyStateHandler = (event: KeyboardEvent) => {
+  handleKeyState = (event: KeyboardEvent) => {
     this.setState({
       pressedKey: `The last pressed key is [${event.key}]`,
     });
@@ -29,9 +29,7 @@ export class App extends Component<{}, State> {
     return (
       <div className="App">
         <p className="App__message">
-          { !pressedKey
-            ? 'Nothing was pressed yet'
-            : pressedKey}
+          { pressedKey || 'Nothing was pressed yet'}
         </p>
       </div>
     );


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends Component<{}, State> {
+  state: State = {
+    pressedKey: 'Nothing was pressed yet',
```
silchyn, line 9: "Nothing was pressed yet" is not a pressed key

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({
+        pressedKey: `The last pressed key is [${event.key}]`,
+      });
+    });
+  }
```
silchyn, line 26: are you sure that works?

silchyn (changes requested):

src/App.tsx
dimoshen: fixed

src/App.tsx
dimoshen: fixed

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          { !pressedKey
+            ? 'Nothing was pressed yet'
+            : pressedKey}
```
naziamoff, line 34: - no need to add this space
- you should checking your booleans from the positive condition, as it is much easier to read
```suggestion
          {pressedKey
            ? pressedKey
            : 'Nothing was pressed yet'}
```

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.keyStateHandler);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.keyStateHandler);
+  }
+
+  keyStateHandler = (event: KeyboardEvent) => {
```
naziamoff, line 20: function names should always start with a verb

naziamoff (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKeyState);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyState);
+  }
+
+  handleKeyState = (event: KeyboardEvent) => {
```
polosanya, line 20: I think `handleKeyPressed` or `handleKeyChange` would be more descriptive

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          { pressedKey || 'Nothing was pressed yet'}
```
polosanya, line 32: Nice solution! It might be also useful to create a variable for "Nothing was pressed" message

polosanya (approved):
Well done!

PR #837, opened 2023-06-08
add task solution

NikaNika12:
https://NikaNika12.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleDocumentClick);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleDocumentClick);
+  }
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+    document.addEventListener('keyup', this.handleDocumentClick);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleDocumentClick);
+  }
+
+  handleDocumentClick = (event: KeyboardEvent) => {
+    if (event.key) {
+      this.setState({ pressedKey: 'The last pressed key is [key]' });
```
TarasHoliuk, line 22: `[key]` is just a string here
you should display `event.key` here

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
TarasHoliuk, line 9: save only `event.key` here (empty string by default)
reason: your value in the state is called `pressedKey` so we have to be consistent and save there exactly described value

use [ternary operator](https://react.dev/learn/conditional-rendering#conditional-ternary-operator--) in `render` method to display `'Nothing was pressed yet'` message or `'The last pressed key is [key]'` message depending on value saved in the state

TarasHoliuk (changes requested):

src/App.tsx
NikaNika12: не знаю правильно ли я тебя поняла ну вот как то так

PR #852, opened 2023-06-09
react_keyboard solution

panch1811:
[DEMO LINK](https://panch1811.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 export class App extends React.Component {
   state = {
-    pressedKey: 'Nothing was pressed yet',
+    pressedKey: '',
   };
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,27 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
```
silchyn, line 5: "Nothing was pressed yet" is not a pressed key

silchyn (changes requested):

src/App.tsx
PANCH1811: @silchyn what u mean? it works how it should due to the task. If not - please, give me more details.
image

src/App.tsx
```diff
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: '',
```
silchyn, line 5: why not `null`?

silchyn (approved):

PR #853, opened 2023-06-09
solution

annholovko:
[DEMO LINK](https://annholovko.github.io/react_keyboard/)


33Orange (changes requested):
Almost done

A few moments to fix

33Orange (approved):
Well done

PR #868, opened 2023-06-11
solution

thienlaoo:
[DEMO LINK](https://thienlaoo.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: KeyboardEvent | undefined;
+};
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    pressedKey: undefined,
+  };
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+    this.setState({ pressedKey: event });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {pressedKey ? `The last pressed key is [${pressedKey.key}]` : 'Nothing was pressed yet'}
```
naziamoff, line 30: ternaries should be multilined

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: KeyboardEvent | undefined;
+};
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    pressedKey: undefined,
```
naziamoff, line 9: null works better if you set the absence of some entity explicitly

naziamoff (approved):
not bad, but consider my comments

PR #876, opened 2023-06-14
Develop

superzuc:
https://superzuc.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,31 +3,34 @@ import React from 'react';
 type Props = {};
 
 type State = {
-  string: string
+  pressedKey: string | null;
 };
 
 export class App extends React.Component<Props, State> {
   state = {
-    string: '',
+    pressedKey: null,
   };
 
   componentDidMount(): void {
-    this.setState({ string: 'Nothing was pressed yet' });
+    document.addEventListener('keyup', this.handleKeyUp);
   }
 
-  componentDidUpdate(): void {
-    document.addEventListener('keyup', (event: KeyboardEvent) => {
-      return (
-        this.setState({ string: `The last pressed key is [${event.key}]` }));
-    });
+  componentWillUnmount(): void {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,34 @@
...
-);
+type Props = {};
+
+type State = {
+  string: string
+};
+
+export class App extends React.Component<Props, State> {
+  state = {
+    string: '',
```
lerastarynets, line 11: it's not a good idea to store a whole string here, it's better to move the string to jsx and only store a value of pressedKey in the state

lerastarynets (changes requested):
Nice start! Left few comments to avoid unexpected behaviour of your code in case user is pressing keys frequently)

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  pressedKey: string | null;
+};
+
+export class App extends React.Component<Props, State> {
+  state = {
+    pressedKey: null,
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyUp);
```
lerastarynets, line 15: Consider using arrow functions in the event listener to automatically bind 'this'.

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {pressedKey ? `The last pressed key is [${pressedKey}]` : 'Nothing was pressed yet'}
```
lerastarynets, line 32:
```suggestion
         pressedKey
      ? `The last pressed key is [${pressedKey}]`
      : 'Nothing was pressed yet';
```

lerastarynets (approved):
Well done with fixing comments) Left one tiny suggestion for the improvement

PR #883, opened 2023-06-16
solution

Kinderwagen:
[DEMO LINK](https://Kinderwagen.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKeyPress);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyPress);
+  }
+
+  handleKeyPress = (event: KeyboardEvent) => {
```
etojeDenys, line 16: It would be better to use a more specific event type 'React.KeyboardEvent' instead of 'KeyboardEvent' to ensure compatibility with React.

etojeDenys (approved):

PR #885, opened 2023-06-19
FedenkoReactKeyboard

VladFedenko:
https://VladFedenko.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,33 @@
-import React from 'react';
+import { Component } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends Component {
+  state = {
+    pressedKey: '',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (evnt: KeyboardEvent) => {
+      this.setState({ pressedKey: evnt.key });
+    });
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', (evnt: KeyboardEvent) => {
```
etojeDenys, line 15: You need to remove the event listener with the same function reference used in `componentDidMount`. Create a class method named `handleKeyUp` and use it as the event listener in both `componentDidMount` and `componentWillUnmount`.

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #887, opened 2023-06-20
solution

BipBobBoop:
[DEMO LINK](https://BipBobBoop.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,29 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    message: null,
+  };
+
+  componentDidMount() {
+    this.setState({ message: 'Nothing was pressed yet' });
+  }
+
+  componentDidUpdate() {
+    document.addEventListener('keyup', (event) => {
+      this.setState({ message: `The last pressed key is [${event.key}]` });
+    });
+  }
+
+  render() {
+    const { message } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {`${message}`}
+        </p>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+  state = {
+    message: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handlerKeyUp);
+  }
+
+  componentDidUpdate() {
+    document.removeEventListener('keyup', this.handlerKeyUp);
```
Anastasiia-Svintsova, line 13: You should remove the listener on the unmount, not on the update. Now the handlerKeyUp works only after the first click. But it should be called on every keyboard press.

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+  }
+
+  handlerKeyUp = (event: KeyboardEvent) => {
+    this.setState({ message: `The last pressed key is [${event.key}]` });
```
Anastasiia-Svintsova, line 17: It's better to save only the `event.key` value in the state because it minimizes storage requirements and simplifies state management.

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+  };
+
+  render() {
+    const { message } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {`${message}`}
```
Anastasiia-Svintsova, line 26:
```suggestion
          {`The last pressed key is [${message}]`}
```

Anastasiia-Svintsova (changes requested):
Hey! Pay attention to my comments, please

src/App.tsx
BipBobBoop: Я все виправив, неуважно прочитав завдання...

Anastasiia-Svintsova (approved):
👍

PR #915, opened 2023-07-17
Solution

adv555:
**[DEMO LINK](https://adv555.github.io/react_keyboard/)**

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,44 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+export class App extends React.Component<{}, State> {
+  state:State = {
+    pressedKey: '',
+  };
+
+  componentDidMount(): void {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,44 @@
...
+  state:State = {
+    pressedKey: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleKeyUp);
```
And678, line 16: This will not remove event listener
https://dev.to/marcostreng/how-to-really-remove-eventlisteners-in-react-3och
https://bobbyhadz.com/blog/react-remove-event-listener

And678 (changes requested):

src/App.tsx
adv555: Thx 😊  as I understand I forgot to bind this, so it could add and remove the same function.
```
 constructor(props){
        super(props);
        this.handleKeyUp = this.handleKeyUp.bind(this); //bind function once
    }
```
Anyway, I rewrote it to the functional component, from the recommendations above

And678 (approved):
Approve :checkered_flag:

PR #942, opened 2023-07-20
solution

Valentyna08:
[DEMO LINK](https://Valentyna08.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.PureComponent {
+  state = {
+    pressedKey: '',
+  };
+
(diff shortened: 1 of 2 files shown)


outofs (approved):
GJ!

src/App.tsx
```diff
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.PureComponent {
```
And678, line 3: PureComponent is a class version of React.memo. No need for it here
```suggestion
export class App extends React.Component {
```

And678 (changes requested):
You are creating new arrow functions in `componentDidMount` and `componentWillUnmount`. Unmount will not work this way, you need to pass `this.handleKeyUp` directly to `addEventListener` and removeEventListener.

anstsot (approved):
Good work, approving:sparkles:

PR #981, opened 2023-07-31
Add task solution

kbekher:
[DEMO LINK](https://kbekher.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,44 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type Props = {
+};
+
+type State = {
+  key: string;
+};
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleKeyUp);
+  }
+
+  handleKeyUp = (event: KeyboardEvent) => {
```
Viktor-Kost, line 20: Yes, it's a good practice to add return type of function as you've done above 
```suggestion
  handleKeyUp = (event: KeyboardEvent): void => {
```

Viktor-Kost (changes requested):
Nice!
just minor fixes needed

OleziO (approved):

PR #987, opened 2023-08-05
Add keys listener

AmadeuszAndroid:
[DEMO LINK](https://amadeuszandroid.github.io/react_keyboard/) and add it to the PR description

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,36 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+const emptyKey = '';
+const keyUp = 'keyup';
+
+export class App extends React.Component {
+  state = {
+    pressedKey: emptyKey,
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+  componentWillUnmount() {
+    document.removeEventListener(keyUp, () => {
+      this.setState({ pressedKey: emptyKey });
+    });
+  }
+
+  render() {
+    const { pressedKey } = this.state;
+
+    const message = pressedKey === emptyKey
```
darokrk, line 26: This line can be simplify:

```
const message = !pressedKey
      ? 'Nothing was pressed yet'
      : `The last pressed key is [${pressedKey}]`;
```

darokrk (approved):
Good job 👏🏽  one suggestion added, you can fix it if you want to but it's not necessary 😃

src/App.tsx
AmadeuszAndroid: I reversed this condition, because is easier to understand.

PR #988, opened 2023-08-07
1.0.0

domkab:
https://domkab.github.io/react_keyboard

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  lastKey: string,
+}
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    lastKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    window.addEventListener('keyup', this.handleKeyUp);
+  }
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+interface State {
+  lastKey: string,
+}
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    lastKey: 'Nothing was pressed yet',
```
Anastasiia-Svintsova, line 9: The initial state should be null. The state name is lastKey but `'Nothing was pressed yet'` doesn't look like a key does it?
```suggestion
    lastKey: null,
```

src/App.tsx
```diff
@@ -1,7 +1,33 @@
...
+
+  handleKeyUp = (event: KeyboardEvent) => {
+    this.setState({ lastKey: `The last pressed key is ${event.key}` });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {`${this.state.lastKey}`}
```
Anastasiia-Svintsova, line 28: Use destructuring for the state object to improve the readability

Anastasiia-Svintsova (changes requested):

lerastarynets (approved):
You're doing a fantastic job, keep going!

PR #997, opened 2023-08-21
completed task

Horizon-git:
[DEMO LINK](https://Horizon-git.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,35 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string,
+};
+// eslint-disable-next-line react/prefer-stateless-function
+export class App extends React.Component {
+  state: State = {
+    key: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKeyPress);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyPress);
+  }
+
+  handleKeyPress = (event: KeyboardEvent) => {
+    this.setState({ key: event.key });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.key
+            ? `The last pressed key is ${this.state.key}`
+            : 'Nothing was pressed yet'}
+        </p>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+
+  handleKeyPress = (event: KeyboardEvent) => {
+    this.setState({ key: event.key });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.key
```
volodymyr-soltys97, line 28: Use a destructuring for `key`

volodymyr-soltys97 (changes requested):
You need to add square brackets
image
image

src/App.tsx
Horizon-git: i fixed key in second commit, but on demo is still completed task of 1st commit. And why I need to desctructing key if its the only variable of state in return jsx?

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string,
+};
+// eslint-disable-next-line react/prefer-stateless-function
+export class App extends React.Component {
```
anastasiiavorobiova, line 7: You should provide the props argument to the React.Component, even if it's empty. Change to 'React.Component<{}, State>'.

src/App.tsx
anastasiiavorobiova: And why I need to desctructing key if its the only variable of state in return jsx?


It makes the code a bit more readable. If there are a lot of values in the state it's better to destructure all values and not repeat `this.state` everywhere

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  };
+
+  render() {
+    const { key } = this.state;
+
+    return (
+      <div className="App">
+        <p className="App__message">
+          {key
+            ? `The last pressed key is [${key}]`
```
VitaliyBondarenko1982, line 31: looks like deployed not actual version of code)
Screenshot from 2023-08-22 17-08-17

VitaliyBondarenko1982 (approved):
GJ!

PR #1001, opened 2023-08-24
solution

Dolnys:
[DEMO LINK](https://dolnys.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,27 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,27 @@
...
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: 'Nothing was pressed yet',
+  };
```
klaudia-klis, line 6: I would set pressedKey state to null, and update it so it's only the key name and I would write condition in rendered JSX that returns the right string depending on pressedKey.

klaudia-klis (approved):
Your solution looks good! 👍🏼

src/App.tsx
choeqq: I would leave it as is, both solutions are fine

choeqq (approved):
All good ✅

PR #1004, opened 2023-08-25
add task solution

szymonpachucki:
[DEMO LINK](https://szymonpachucki.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,45 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    wasKeyPressed: false,
+    keyPressed: null,
+  };
+
+  componentDidMount() {
+    window.addEventListener('keydown', this.handleKeyPress);
+  }
+
+  componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


cardinal1991 (approved):
Looks good to me, you can handle Space{' '}press with some description eventually.

src/App.tsx
```diff
@@ -1,7 +1,45 @@
...
+  render() {
+    const { wasKeyPressed, keyPressed } = this.state;
+
+    if (wasKeyPressed) {
+      return (
+        <div className="App">
+          <p className="App__message">
+            The last pressed key is
+            {' '}
+            {keyPressed}
```
JoyCoffeeAddict, line 33: You can use string templates to make it more readable

src/App.tsx
```diff
@@ -1,7 +1,45 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    wasKeyPressed: false,
```
JoyCoffeeAddict, line 5: Can you think what is the value of `keyPressed` when `wasKeyPressed === false`? That should point you into thinking, if you need this additional state property

JoyCoffeeAddict (changes requested):

choeqq (approved):

PR #1007, opened 2023-08-25
test solution

cardinal1991:
[DEMO LINK](https://cardinal1991.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,55 @@
 import React from 'react';
+import './index.scss';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  key: string | null;
+};
+
+export class App extends React.Component<{}, State> {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,55 @@
...
+
+  render() {
+    const { key } = this.state;
+    let content;
+
+    if (!key) {
+      content = (
+        <>
+          Nothing was pressed yet
+        </>
```
choeqq, line 37: you can assign contetn to string instead of the whole ReactElement

src/App.tsx
```diff
@@ -1,7 +1,55 @@
...
+      );
+    } else {
+      content = (
+        <>
+          {`The last pressed key is [${key}]`}
+        </>
```
choeqq, line 43: you can assign contetn to string instead of the whole ReactElement

choeqq (changes requested):
please consider rendering text using ternary operator

choeqq (approved):

PR #1023, opened 2023-08-28
Develop

Matylda-Lewandowska:
[DEMO LINK](https://matylda-lewandowska.github.io/react_keyboard/)
task solution - keyboard

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,39 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    pressedKey: 'Nothing was pressed yet',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyUp);
+  }
+
+  handleKeyUp = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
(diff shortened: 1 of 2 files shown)


Sancz0pansa (approved):
Looks fine 👍

src/App.tsx
```diff
@@ -1,7 +1,39 @@
...
+    let appMessage = <p className="App__message">Nothing was pressed yet</p>;
+
+    if (pressedKey !== 'Nothing was pressed yet') {
+      appMessage = <p className="App__message">{`The last pressed key is [${pressedKey}]`}</p>;
+    }
+
+    return (
+      <div className="App">
+        {appMessage}
+      </div>
```
mvjl000, line 36: This works but having something like ternary operator in the jsx would be easier to mantain.
The only thing changing here is the text, paragraph tag stays the same.

For example - If you wanted to make some change to p tag className, you'd have to do it twice ;)

How about these?
```suggestion
    const appMessage = pressedKey !== 'Nothing was pressed yet' ? `The last pressed key is [${pressedKey}]` : pressedKey;

    return (
      <div className="App">
        <p>{appMessage}</p>
      </div>
```

mvjl000: Or an even cleaner solution:
Since initial state of pressedKey is 'Nothing was pressed yet' (doesn't look like a key, I suggest renaming that), maybe you could put there the entire phrase on the keyboard event?

So the inside of handleKeyUp would look like this: `this.setState({ pressedKey: `The last pressed key is [${event.key}]` });`

And then in the jsx you could simply put the state value in the p tag, no if statements needed

mvjl000 (changes requested):
Minor changes / suggestions ;)

Matylda-Lewandowska:
I hope now is better! :)

asakevych (approved):
Great job!!! 🚀 🔥 👏 👏

PR #1034, opened 2023-08-31
keyboard

DenisGurskiy:
https://DenisGurskiy.github.io/react_keyboard/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,41 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    key: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyPressed);
+  }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+  handleKeyPressed = (event: KeyboardEvent) => {
+    this.setState({
+      key: event.key,
+    });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        {this.state.key !== ''
```
nazarmatsevych, line 25:
```suggestion
        {this.state.key
```

Same but a bit shorter

src/App.tsx
```diff
@@ -1,7 +1,41 @@
...
+            <p className="App__message">
+              The last pressed key is [
+              {this.state.key}
+              ]
+            </p>
+          )
+          : (
+            <p className="App__message">
+              Nothing was pressed yet
+            </p>
```
nazarmatsevych, line 36: You can also use ternary just for content and no need to duplicate p tag with class ` App__message`

nazarmatsevych (changes requested):

volodymyr-soltys97 (approved):
Well done 🔥

PR #1038, opened 2023-09-02
added react_keyboard

Ilya-Osoben:
[DEMO LINK](https://Ilya-Osoben.github.io/react_keyboard/)

### DIFF
--- a/src/components/Message/Message.tsx
+++ b/src/components/Message/Message.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+
+type Props = {};
+
+type State = {
+  query: string;
+};
+
+export class Message extends React.Component<Props, State> {
+  state: State = {
+    query: '',
+  };
+
+  componentDidMount(): void {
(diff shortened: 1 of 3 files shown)


src/components/Message/Message.tsx
```diff
@@ -0,0 +1,37 @@
...
+  componentDidMount(): void {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ query: event.key });
+    });
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ query: event.key });
+    });
```
etojeDenys, line 23: it's two different functions here, so remove event listener will not work as you can expect, you should write that func in a separate function

etojeDenys (changes requested):

VitaliyBodnar (approved):
Good job!

PR #1048, opened 2023-09-07
Implemented logic for printing a last pressed button

davinnchii:
[DEMO LINK](https://davinnchii.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,27 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export const App: React.FC = () => {
+  const [key, setKey] = useState('');
+
+  const handleKeyPress = (event: KeyboardEvent) => {
+    setKey(event.key);
+  };
+
+  useEffect(() => {
+    document.addEventListener('keyup', handleKeyPress);
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,27 @@
...
+    setKey(event.key);
+  };
+
+  useEffect(() => {
+    document.addEventListener('keyup', handleKeyPress);
+
+    return () => {
+      document.removeEventListener('keyup', handleKeyPress);
+    };
+  }, [key]);
```
DenisChakhalian, line 16: image
but you used componentDidUpdate

src/App.tsx
```diff
@@ -1,7 +1,27 @@
...
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export const App: React.FC = () => {
+  const [key, setKey] = useState('');
+
+  const handleKeyPress = (event: KeyboardEvent) => {
+    setKey(event.key);
+  };
+
```
DenisChakhalian, line 9: it may be better to write a constant in useEffect as it will be re-rendered after each render

DenisChakhalian (approved):
good job, but there are some things that can be fixed if you want

src/App.tsx
davinnchii: @DenisChakhalian good point, forgot to fix it, thanks!

roman-mirzoian (approved):
Well done 👍

PR #1049, opened 2023-09-08
completed task react_keyboard

DenisChakhalian:
[DEMO LINK](https://denischakhalian.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,27 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export const App: React.FC = () => {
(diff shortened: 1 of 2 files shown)


alexander-ignatow (changes requested):
I really love your solution, and it is how I would do it.

But per task you need a class component, not functional:
> Make the App a class component with pressedKey in the state.

src/App.tsx
```diff
@@ -1,7 +1,27 @@
...
+    const handleDocumentKeyup = (event: KeyboardEvent) => {
+      setPressedKey(event.key);
+    };
+
+    document.addEventListener('keyup', handleDocumentKeyup);
+
+    return () => {
+      document.removeEventListener('keyup', handleDocumentKeyup);
+    };
+  }, []);
```
alexander-ignatow, line 16: It would have nearly zero effect, but technically you need to have your `setPressedKey` as a dependency to the effect hook

roman-mirzoian (approved):
Well done 👍

PR #1051, opened 2023-09-09
Task solution

yuliiaant:
[DEMO LINK](https://yuliiaant.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,55 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.PureComponent<State> {
+  state: State = {
+    pressedKey: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeySet);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleKeySet);
+  }
+
+  handleKeySet = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,7 +1,55 @@
...
+            className="App__message"
+          >
+            The last pressed key is [
+            {pressedKey}
+            ]
+            <button
+              type="button"
+              className="delete"
+              aria-label="close"
+              onClick={() => this.handleKeyUnset()}
```
rostyslav48, line 44: I am not sure whether you had to add this button (it is not mentioned in the task), but it works, so let it be)

src/App.tsx
```diff
@@ -1,7 +1,55 @@
...
+  render() {
+    const { pressedKey } = this.state;
+
+    return (
+      <div className="App">
+        {this.state.pressedKey ? (
+          <p
+            className="App__message"
+          >
+            The last pressed key is [
```
rostyslav48, line 37: It is better to use interpolation instead of writing it in this way. Now you are connecting some HTML text with JS variable, as far as I know it is not a good practice)
`{`The last pressed key is ${this.state.pressedKey}`}`

rostyslav48 (approved):
Great Job😁

src/App.tsx
```diff
@@ -1,7 +1,53 @@
...
+        {this.state.pressedKey ? (
+          <p
+            className="App__message"
+          >
+            {`The last pressed key is [${pressedKey}]`}
+            <button
+              type="button"
+              className="delete"
+              aria-label="close"
+              onClick={() => this.handleKeyUnset()}
```
roman-mirzoian, line 42:
```suggestion
              onClick={this.handleKeyUnset}
```

roman-mirzoian, line 42: Interesting feature, great job!

roman-mirzoian (approved):
Well done 👍

PR #1052, opened 2023-09-09
add key event listener with class component

rostyslav48:
[DEMO_LINK](https://rostyslav48.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,19 +8,17 @@ export class App extends React.Component<{}, State> {
     pressedKey: '',
   };
 
-  private NO_KEY_PRESSED = 'Nothing was pressed yet';
-
   componentDidMount(): void {
-    document.addEventListener('keyup', (event: KeyboardEvent) => {
-      this.setState({ pressedKey: event.key });
-    });
+    document.addEventListener('keyup', this.handleKeyPress);
   }
 
   componentWillUnmount(): void {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
-);
+type State = {
+  pressedKey: string,
+};
+export class App extends React.Component<{}, State> {
+  state: State = {
+    pressedKey: '',
+  };
+
+  private NO_KEY_PRESSED = 'Nothing was pressed yet';
```
Koliras, line 11: I don't think you need to create variable for this message since you use it only once

src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({ pressedKey: event.key });
+    });
+  }
```
Koliras, line 23: I think you should make `this.setState({ pressedKey: event.key });` a function, since you use it twice

Koliras (changes requested):

src/App.tsx
alexander-ignatow: IMO it would be an overkill :)

alexander-ignatow (approved):
LGTM 👍

PR #1055, opened 2023-09-11
Solved task

AndriiKuz:
[DEMO LINK](https://AndriiKuz.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,43 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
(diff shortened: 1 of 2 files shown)


vserhiichuk (approved):
It looks good

src/App.tsx
```diff
@@ -1,7 +1,43 @@
...
+          ? (
+            <p className="App__message">
+              {`The last pressed key is [${pressedKey}]`}
+            </p>
+          )
+          : (
+            <p className="App__message">
+              Nothing was pressed yet
+            </p>
+          )}
```
roman-mirzoian, line 39: You don't need to write the same <p> tags twice - you should write one tag and add content depending on the condition.

roman-mirzoian (changes requested):

PR #1070, opened 2023-09-12
add keyup logic to app

Denys14Samoilenko:
[DEMO LINK](https://Denys14Samoilenko.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,37 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string
+};
+
+export class App extends React.Component<{}, State> {
+  state: State = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  }
+
+  handleKey = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: `${event.key}` });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <a href="/" className="App__message">
```
Taras2705, line 27: good job but why did you change 'p' to  'a'?

Taras2705 (approved):
Good job !

src/App.tsx
roman-mirzoian: I have the same question: after clicking on this link, the site shows an error, so it should be fixed.

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done 👍

PR #1072, opened 2023-09-12
implement keyboard logic using life methods in class component

serkrops:
[demo link](https://serkrops.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,44 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type Props = {};
+type State = {
+  text: string;
+};
+
+export class App extends React.Component<Props, State> {
+  state = {
+    text: '',
+  };
+
+  componentDidMount(): void {
+    this.handleState();
+  }
+
+  componentDidUpdate(): void {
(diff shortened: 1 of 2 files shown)


voievodik (changes requested):
Looks good! Please pay attention on requirements in task.

src/App.tsx
```diff
@@ -1,7 +1,44 @@
...
+    text: '',
+  };
+
+  componentDidMount(): void {
+    this.handleState();
+  }
+
+  componentDidUpdate(): void {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
```
voievodik, line 19: [checklist] use componentDidMount to add keyup handler. 

You added event listener for componentDidUpdate, but you don't need to use componentDidUpdate in this situation.

src/App.tsx
```diff
@@ -1,7 +1,44 @@
...
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleKeyUp);
+  }
+
+  handleState = () => {
+    this.setState({
+      text: 'Nothing was pressed yet',
+    });
+  };
```
voievodik, line 29: [checklist] Make the App a class component with pressedKey in the state.

Also you need keep in state current key, but not a full sentence. You can use conditional rendering to display the desired text

voievodik (approved):
Good job!

roman-mirzoian (approved):
Well done 👍

PR #1074, opened 2023-09-12
evetntKeyboard

voronine:
[DEMO LINK](https://voronine.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,46 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    pressKey: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyPress);
+  }
+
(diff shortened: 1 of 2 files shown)


KatOlista (approved):
good job!

src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
+            The last pressed key is
+            [
+            {pressKey}
+            ]
+          </p>
+        ) : (
+          <p className="App__message">
+            Nothing was pressed yet
+          </p>
+        )}
```
roman-mirzoian, line 42: You do not need to write two identical <p> tags - it is enough to write this tag once and move the condition for the message inside it.

roman-mirzoian (changes requested):

src/App.tsx
voronine: I agree!

src/App.tsx
```diff
@@ -1,7 +1,40 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressKey: string;
```
Moroz-Dmytro, line 4:
```suggestion
  pressedKey: string;
```

Moroz-Dmytro (approved):

PR #1078, opened 2023-09-12
added logic for the react keyboard task using lifecycle methods in cl…

irahos:
[DEMO LINK](https://irahos.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleKeyUp);
+  }
+
+  handleKeyUp = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    return (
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component {
```
desevoker, line 3:
```suggestion
import { PureComponent } from 'react';

type State = {
  pressedKey: string;
};

export class App extends PureComponent<{}, State> {
```
What if we add some typization?
And also change the component type so it won't rerender on each keypress.

src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+  handleKeyUp = (event: KeyboardEvent) => {
+    this.setState({ pressedKey: event.key });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <p className="App__message">
+          {this.state.pressedKey
+            ? `The last pressed key is [${this.state.pressedKey}]`
```
desevoker, line 25:
```suggestion
  render() {
    const { pressedKey } = this.state;

    return (
      <div className="App">
        <p className="App__message">
          {pressedKey
            ? `The last pressed key is [${pressedKey}]`
```
And a touch of destructuring.

desevoker (approved):
Goob job! 🚀

Left a couple of comments for you, check them out.

src/App.tsx
```diff
@@ -1,7 +1,31 @@
 import React from 'react';
```
desevoker, line 1:
```suggestion
import { PureComponent } from 'react';
```
Missed this line, it should be like this.

PR #1090, opened 2023-09-12
done with FC and CC

maxonshev:
https://maxonshev.github.io/react_keyboard/

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -2,8 +2,12 @@ import ReactDOM from 'react-dom';
 import { App } from './App';
 
 import './index.scss';
+import { App2 } from './App2';
 
 ReactDOM.render(
-  <App />,
+  <>
+    <App />
+    <App2 />
+  </>,
   document.getElementById('root'),
 );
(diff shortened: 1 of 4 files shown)


src/index.tsx
```diff
@@ -2,8 +2,12 @@ import ReactDOM from 'react-dom';
...
 
 import './index.scss';
+import { App2 } from './App2';
 
 ReactDOM.render(
-  <App />,
+  <>
+    <App />
+    <App2 />
+  </>,
```
Viktor-Kost, line 11: it's nice that you used national colors) yet your code duplicate each other. If you'd like to use two paragraphs with different color you can use it inside one App component, just use two `p`

Viktor-Kost (changes requested):

src/index.tsx
maxonshev: For practice, I implemented App using a Class Component and App2 using a Functional Component.

etojeDenys (approved):
good job

Viktor-Kost (approved):

PR #2039, opened 2025-03-23
solution

dmytroKolesnikov52:
[DEMO LINK](https://Mazafakerqq.github.io/react_keyboard/)


ENigmusSS (changes requested):
Make the `App` a class component with `pressedKey` in the `state`. ТЗ читати треба)

PR #2044, opened 2025-03-25
Solution

Olehgoon093:
[DEMO LINK](https://Olehgoon093.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,40 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,7 +1,40 @@
...
+export class App extends React.Component<{}, State> {
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount(): void {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      // eslint-disable-next-line no-console
+      console.log(event.key);
+      this.setState({ pressedKey: `[${event.key}]` });
```
maxbilfed: You need to store your function in a class method, and then use it here. Something like:
`handleKeypress = (event: KeyboardEvent) => {
      this.setState({ pressedKey: `[${event.key}]` });
  };`
  And
  `componentDidMount(): void {
    document.addEventListener('keyup', this.handleKeypress);
  }`
  Same with `componentWillUnmount()`
  
  Also, you can just remove the console.log lines

maxbilfed (changes requested):
Almost there

PR #2116, opened 2025-05-20
add solution

alexmotalex:
- [DEMO LINK](https://alexmotalex.github.io/react_keyboard/)


alexmotalex:
@mateacademy-ai-mentor

Hi there. You say that there is a typo in the name of the lifecycle method: it should be componentWillUnmount, not componentWillUnmount. But it is identical, they are the same methods

PR #2131, opened 2025-06-02
Solution

Ivancvnts:
[DEMO LINK](https://Ivancvnts.github.io/react_keyboard/)


Ivancvnts:
@mateacademy-ai-mentor a change has been requested but it doesnt make any sense since its requesting to change to what already is:

There is a typo in the lifecycle method name: it should be componentWillUnmount, not componentWillUnmount. This prevents the cleanup code from running, which is a critical issue according to React's lifecycle method requirements.

PR #2172, opened 2025-06-09
add task solution

Khmara-Danylo:
[DEMO LINK](https://Khmara-Danylo.github.io/react_keyboard/)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleDocumentKeyUp);
+  }
+
+  render() {
+    return (
+      <div className="App">
+        {this.state.pressedKey.length === 0 ? (
+          <p className="App__message">Nothing was pressed yet</p>
```
BudnikOleksii, line 28: Element is the same `<p className="App__message">` in both cases, only content is different. So change only it

BudnikOleksii (approved):

PR #2181, opened 2025-06-17
add task solution

ihorkrasnozhon:
- [DEMO LINK](https://ihorkrasnozhon.github.io/react_keyboard/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,31 +1,27 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 
-export class App extends React.Component<{}, { pressedKey: null | string }> {
-  state = {
-    pressedKey: null,
-  };
+export const App: React.FC = () => {
+  const [pressedKey, setPressedKey] = useState<string | null>(null);
 
-  componentDidMount() {
-    document.addEventListener('keyup', this.handleKeyUp);
-  }
+  useEffect(() => {
+    const handleKeyUp = (event: KeyboardEvent) => {
+      setPressedKey(event.key);
+    };
 
-  componentWillUnmount() {
-    document.removeEventListener('keyup', this.handleKeyUp);
-  }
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,31 @@
...
+    pressedKey: null,
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.handleKeyUp);
+  }
```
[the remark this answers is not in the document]
ihorkrasnozhon: "it should be !componentWillUnmount! , not !componentWillUnmount! " - similar

src/App.tsx
```diff
@@ -1,7 +1,31 @@
 import React from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export class App extends React.Component<{}, { pressedKey: null | string }> {
```
danon321: Why u change from hooks approach to the class approach?
Before uv got export const App: React.FC = () => (
and now u have class App extends React.Component<{},

Keep it in the first one cuz its more modern aproach

danon321 (changes requested):

src/App.tsx
```diff
@@ -1,7 +1,27 @@
...
+
+    return () => {
+      document.removeEventListener('keyup', handleKeyUp);
+    };
+  }, []);
+
+  return (
+    <div className="App">
+      <p className="App__message">
+        {pressedKey === null
```
natalia-klonowska: porównanie z null jest zbędne ponieważ wystarczy sprawdzić czy pressedKey ma wartość :)

natalia-klonowska (changes requested):

natalia-klonowska (approved):

PR #2257, opened 2025-08-18
solution

tavisua:


etojeDenys (changes requested):
fix the demo link and then request the review again
image

etojeDenys (approved):
lgtm

PR #2282, opened 2025-08-20
Develop

IlchenkoDima:
https://IlchenkoDima.github.io/react_keyboard


horidor (approved):
Good job! Code correctly handles input from DOM model while utilizing class component.

PR #2296, opened 2025-08-25
solution

vladsulyma:


src/App.tsx
```diff
@@ -1,7 +1,27 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <p className="App__message">The last pressed key is [Enter]</p>
-  </div>
-);
+export const App: React.FC = () => {
```
derpedcatto: I think this task specified the usage of Class component instead of Functional, but if this will pass tests then whatever

derpedcatto (approved):

PR #2301, opened 2025-08-29
Solution

CyberDogFK:
[DEMO LINK](https://cyberdogfk.github.io/react_keyboard/)


src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+  state = {
+    pressedKey: '',
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (event: KeyboardEvent) => {
+      this.setState({
+        pressedKey: event.key,
+      });
+    });
```
BudnikOleksii, line 17: Create a handler instead of passing anonim function to event listener directly so you can remove it on unmount

src/App.tsx
```diff
@@ -1,7 +1,35 @@
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', () => {});
```
BudnikOleksii, line 21: You should remove a handler
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

BudnikOleksii (changes requested):
Good work 👍 
Check comments

PR #2368, opened 2025-10-21
initial commit

Pobedasg:


anastasia-kravchuk (changes requested):
Please remove console.log and 'as any' because that's why you fail tests and optional rename lastPressedKey to just a pressedKey for consistency with the task and cleaner naming

PR #2462, opened 2026-01-21
added solution

Demtar23:
[DEMO LINK](https://Demtar23.github.io/react_keyboard/)


src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+type State = {
+  pressedKey: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    pressedKey: '',
+  };
+
+  handleClick = (event: KeyboardEvent) => {
```
dmytro-halieba, line 12: I would rename this function to `handleKeyUp` because `handleClick` is misleading - it suggests a mouse click event, not a keyboard event.

src/App.tsx
```diff
@@ -1,7 +1,37 @@
...
+  }
+
+  componentWillUnmount(): void {
+    document.removeEventListener('keyup', this.handleClick);
+  }
+
+  render() {
+    return (
+      <div className="App">
+        {this.state.pressedKey.length === 0 ? (
```
dmytro-halieba, line 27: Also, you can simplify the condition from `this.state.pressedKey.length === 0` to just `!pressedKey` after destructuring. It's shorter and more idiomatic.

dmytro-halieba (approved):
Great job!

PR #2467, opened 2026-01-21
Make the App a class component with pressedKey in the state

NastyBaster:


Staskovych (approved):
good job, I don't see any problems

PR #2468, opened 2026-01-21
Solution

mark-chep:


NastyBaster (approved):
Since we are strictly limited by references in this task, this is the only correct solution.
🚀 Great job!

PR #2592, opened 2026-05-05
implement solution

whomngmnt:
[DEMO LINK]  https://whomngmnt.github.io/react_keyboard/


timurradkevic (approved):
👍, but you should add console.log to the handler, and it would be better to rename it to handlerKeyup

PR #2595, opened 2026-05-06
Solution

OlehYuriev:
[DEMO LINK](https://OlehYuriev.github.io/react_keyboard/)


timurradkevic (approved):
👍 but you should add a log in the keyupHandler

PR #2603, opened 2026-05-12
add task solution

rlysunets:
[DEMO LINK](https://rlysunets.github.io/react_keyboard/)


linska (changes requested):
The solution works, but the task requires a class component with **componentDidMount** and **componentWillUnmount**.
Right now it is implemented with hooks and useEffect, so it doesn't fully follow the task requirements.

linska (approved):
🔥 good job

PR #2605, opened 2026-05-12
Solution

Misharkus:
[DEMO LINK](https://Misharkus.github.io/react_keyboard/)


Evgeniy45 (changes requested):
According to the task, you need do like this:
"document.addEventListener('keyup', (event: KeyboardEvent)"

But your solution is:
"window.addEventListener('keydown', handleKeyDown)"

Please fix this ;)

PR #2619, opened 2026-05-16
solution

Dazavrrr:
- [DEMO LINK](https://Dazavrrr.github.io/react_keyboard/)


Tysiachnyk (approved):
Hello, good work!
Everything works properly!

Here are some thoughts on the code:

1) There is a possible chance that Luke may not approve the solution because of the missing console.log. Tanya mentioned this in the Q&A chat.

If he doesn’t approve it, you can return the log and disable the linter rule above it so it looks like this:

```
handleKeyUp = (event: KeyboardEvent) => {
  // eslint-disable-next-line no-console
  console.log(event.key);
};
```

2) In the conditional rendering part:
`{this.state.pressedKey ? ...}`
you can simply use pressedKey, since you already destructured it above.

3) It would be better to explicitly pass props and state generics to React.Component, like this
`React.Component<{}, State>`
although in this case it’s not critical.
In class components, React.Component accepts two generic parameters:
`React.Component<Props, State>`
If there are no props, you must pass {} as the first argument, otherwise State will end up in the props position.

Continue in the same spirit 👍

dothearts (approved):

PR #2623, opened 2026-05-17
Solution

lilp3ka:
https://lilp3ka.github.io/react_keyboard/


Viqsol (approved):
Nice job! But next time you can use void type for both componentDidMount() and componentWillUnmount() methods.
Also you need use valid link for DEMO:)

PR #2673, opened 2026-07-16
Solution: React Keyboard task 16072026 1744

sistsayao-hue:
DEMOLINK: https://sistsayao-hue.github.io/react_keyboard/


raulriato (changes requested):
Your `DEMO LINK` is not working properly.

It seems you deployed the application before `pushing` your code.

You need to fix it, so we can review your code.

